PR #8, opened 2023-12-19
Prevent preview clicks (#6)

### LINKED ISSUE #6: Prevent preview clicks
As a content creator, I would like links in the preview pane to not work so that I do not browse away from my previewed page.

karlpauls:
Might be a bit heavy handed but I guess you could just disable pointer events.

Fix #6


auniverseaway (approved):
I think this is great. 

I'm a bit shocked this works. I was thinking we would have to inject CSS on the consumer to disable anchors. I think we ship this for now until someone complains.

This would mean that there's no clicking anywhere and that's probably exactly what we want to do.

@davidnuescheler can you think of a scenario where we would want people clicking into the preview?

PR #9, opened 2023-12-19
Update the preview with the editor content when it is available (#7)

### LINKED ISSUE #7: Preview has no body until first content edit if document has not been previewed
When creating a new document, the preview is empty until the document is edited at least one time. 

https://github.com/auniverseaway/da/assets/410610/8541db29-e4a6-4c29-8298-c6083b0c8e28

karlpauls:
Might not be a good idea, but I think one way to fix #7 is to just always initially send the editor content to the preview. Granted, in the normal case, the content is previewed so the preview will see it without that updated and this just adds unnecessary noise - I'll make it a draft just in case.

### DIFF
--- a/blocks/edit/prose/index.js
+++ b/blocks/edit/prose/index.js
@@ -36,20 +36,25 @@ function dispatchTransaction(transaction) {
   view.updateState(newState)
 }
 
+function setPreviewBody(daPreview, proseEl) {
+  const clone = proseEl.cloneNode(true);
+  const body = prose2aem(clone);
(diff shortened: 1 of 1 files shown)


auniverseaway (approved):
This seems right to me. I struggled with how to handle the idea that on first view, the preview is basically a 404. Having it send on first run when it's available passes my expectations.

auniverseaway:
And I think 404 is mostly correct.... we don't want to store docs in content bus unless they hit preview at least once.

Issue #12, opened 2023-12-19
Create document pop-up blocked in Safari


karlpauls:
Safari is blocking the window.open on a create new document.

PR #13, opened 2023-12-19
Fix pop-up window on create new document in safari

karlpauls:
I think fixing #12 is as simple as just doing the window.open before awaiting the document creation in da.

Fix: #12

### DIFF
--- a/blocks/browse/da-browse/da-browse.js
+++ b/blocks/browse/da-browse/da-browse.js
@@ -80,9 +80,9 @@ export default class DaBrowse extends LitElement {
       const ext = 'html';
       path += `.${ext}`;
       const blob = new Blob([''], { type: 'text/html' });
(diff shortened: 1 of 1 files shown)


auniverseaway (commented):
The only thing is that saving the empty doc to DA will now happen **_after_** the window opens. This will technically create a race condition... it probably won't matter because the doc is empty anyway... but the window will open and try to fetch the doc that may not exist yet.

I'd like to have a more robust flag for calling out a popup was blocked. From my initial research, `window.open` when successful returns _something_. If it doesn't return, there's a high likelihood the popup was blocked.

It would be nice to have some kind of notification telling people what is going on.

karlpauls:
I'm not sure it matters so as it is fine to just go to the none existing url :-)

karlpauls:
Closed in favor of https://github.com/auniverseaway/da/pull/15

PR #47, opened 2024-01-22
Make the collab server configurable including which da-admin it shoul…

karlpauls:


auniverseaway (approved):
Once this is merged I'll do a single PR for all constants to be the correct CASE.

These IIFEs are great so they execute once during the module import and then the constant is available for re-use in all other places.

auniverseaway:
If you wanted, you could get a head start and change your case to...

```
const COLLAB_ORIGIN = ...
```

as part of this pr. Or you can talk me out of doing this case fix. :-)

blocks/edit/prose/index.js
```diff
@@ -80,8 +81,8 @@ export default function initProse({ editor, path }) {
 
   const ydoc = new Y.Doc();
 
-  const server = 'wss://collab.da.live';
-  const roomName = `https://admin.da.live${new URL(path).pathname}`;
+  const server = `${collabOrigin}`;
```
auniverseaway, line 84: nit:
```suggestion
  const server = collabOrigin;
```

Or just use collabOrigin for the roomName prefix.

blocks/edit/prose/index.js
```diff
@@ -80,8 +81,8 @@ export default function initProse({ editor, path }) {
...
+  const roomName = `${origin}${new URL(path).pathname}`;
```
auniverseaway, line 85: nit:
```suggestion
  const roomName = `${collabOrigin}${new URL(path).pathname}`;
```

auniverseaway: Ooooh. I see. I misread.

PR #63, opened 2024-02-07
Setup linting and adjust accordingly

karlpauls:


auniverseaway (approved):
There's still more work to do, but I think this is a great start. Lots of ignores we should fix and polish over time.

Even reviewing this PR, I'm struck by how much code is already in DA.

PR #64, opened 2024-02-12
Don't request content from da-admin

karlpauls:


auniverseaway (approved):

auniverseaway:
FWIW: this dies if IMS dies. So for branches (like this one) that go beyond our IMS domain character limit, edit won't work.

I think there's a try / catch in daFetch that may provide some guidance on how to handle this a little more gracefully.

karlpauls:
> FWIW: this dies if IMS dies. So for branches (like this one) that go beyond our IMS domain character limit, edit won't work.
> 
> I think there's a try / catch in daFetch that may provide some guidance on how to handle this a little more gracefully.

@auniverseaway, good point - I created #66 to address this.

PR #66, opened 2024-02-14
Unify ims init and fix error handling

karlpauls:
We now load ims in several places and don't handle the case where we get an error (due to the domain size limit). This PR fixes this.

### DIFF
--- a/blocks/edit/da-editor/da-editor.js
+++ b/blocks/edit/da-editor/da-editor.js
@@ -1,12 +1,9 @@
 import { LitElement, html } from '../../../deps/lit/lit-core.min.js';
 import initProse from '../prose/index.js';
 import getSheet from '../../shared/sheet.js';
+import { initIms } from '../../shared/utils.js';
 
 const sheet = await getSheet('/blocks/edit/da-editor/da-editor.css');
-
-// Milo Imports
-const { getLibs } = await import('../../../scripts/utils.js');
-const { loadIms } = await import(`${getLibs()}/utils/utils.js`);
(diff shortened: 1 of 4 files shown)


blocks/edit/da-editor/da-editor.js
```diff
@@ -18,9 +15,11 @@ export default class DaEditor extends LitElement {
     super.connectedCallback();
     this.shadowRoot.adoptedStyleSheets = [sheet];
     this.shadowRoot.createRange = () => document.createRange();
-    loadIms().then(() => {
-      this._imsLoaded = true;
-    });
+    if (!this._imsLoaded) {
```
auniverseaway, line 18: Is this necessary? Connected callback should only be called when the component is attached to the DOM... which would mean this is never true.

https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#:~:text=called%20each%20time%20the%20element%20is%20added%20to%20the%20document.

auniverseaway (approved):
I think our IMS handling could still use improvement / be more consistent, and I'm not sure this solves the issues I _think_ we have, but it's a step in the right direction.

Merge away when you're ready.

blocks/edit/da-editor/da-editor.js
karlpauls: Good point, removed it.

karlpauls:
Yeah, it at a minimum does not solve the issue where the domain is too long - but at least it does not make use fail in that case anymore.

Issue #84, opened 2024-03-19
Create E2E tests


bosschaert:
We should have E2E tests to check the da.live behaviour.

PR #85, opened 2024-03-19
GH-84 - Some initial e2e tests

bosschaert:
## Description

Use Playwright to create an initial set of E2E tests.

This PR contributes the setup for Playwright which is in the `e2e_test` subdirectory.
Tests can be run from the root using `npm run test:e2e` or using the various `npm run test:...` scripts in the `e2e_test` directory.
Normally the tests are run on Chrome, Firefox and Webkit.

This PR contributes 3 tests:
* A simple test that just obtains the home page and checks that some text on it appears
* A test that updates a document and then loads it in a different browser window to check that the updates were persisted
* A test that creates a new document and then deletes it

This is only the start of the E2E test suite. Hopefully many more test cases will follow.

## Related Issue

Contributes to #84

## Motivation and Context

E2E tests are important to ensure product quality.

## How Has This Been Tested?

It is a test 😉 

## Types of changes

## Checklist:

- [ ] All new and existing tests passed - see note.


bosschaert:
Note that one of the tests fails on webkit. This is because the checkbox widget is not tab-selectable there. I left the test failure in there for now as this might be something we really want to fix?

bosschaert:
I've limited the runs to be on Chromium and Firefox for now until the checkbox selection issue on Webkit is fixed. That way the playwright tests pass.

e2e_test/tests/editor.spec.js
```diff
@@ -0,0 +1,93 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+/* eslint-disable import/no-unresolved */
+const { test, expect } = require('@playwright/test');
+
+test('Get Main Page', async ({ page }) => {
+  await page.goto('https://da.live/');
```
auniverseaway, line 16: We want this on the PR URL, right?

We did this in our playwright project... we grabbed the incoming branch name and constructed the URL to run the tests.

auniverseaway:
What do you think about...
```
test
  ut
  e2e
```

Not sure if `ut` is a good name, but it would be nice to not take up a top level folder. Seems like all tests could live in the same folder.

bosschaert:
> What do you think about...
> 
> ```
> test
>   ut
>   e2e
> ```
> 
> Not sure if `ut` is a good name, but it would be nice to not take up a top level folder. Seems like all tests could live in the same folder.

Good point, I relocated the tests now as follows (so used `unit` for the unit tests):

```
test
  unit
  e2e
```

e2e_test/tests/editor.spec.js
bosschaert: Of course, thanks. I've done that now, the test URL is echoed so you can verify it: https://github.com/adobe/da-live/actions/runs/8419149350/job/23051143836?pr=85#step:6:9

bosschaert:
@auniverseaway I think I addressed your concerns, let me know if you need any further changes.

auniverseaway (approved):
One last small thing I'd recommend changing.

auniverseaway:
I don't think I have a strong opinion on this, but it's an interesting observation:

Having another package.json in a sub-folder puts us squarely in `monorepo` territory. We know Playwright is a bit heavy from an npm package perspective, so this makes a ton of sense... if you do `npm install` on the root, you only get what you absolutely need. It's worth calling out that this is an outlier right now since WTR is not in a separate package. I can see someone preferring a consistent strategy for `package.json`...

```
package.json (linting, app dependencies, etc.)
  test
    unit
      package.json (Web Test Runner)
    e2e
      package.json (Playwright)
```

test/e2e/tests/editor.spec.js
```diff
@@ -0,0 +1,97 @@
+/*
```
auniverseaway, line 1: Sorry. Keep finding little things...

Would it make sense to call these files the same as their associated views?

```
/       >>> browse
/edit   >>> edit
/sheet  >>> sheet
/media  >>> media
/config >>> config
```

auniverseaway: Not sure if it will scale this way forever. 😄

auniverseaway (approved):

PR #102, opened 2024-04-09
Image Alt Text

chrischrischris:
## Description

Add ability to set Alt text on an image.

Also:
* Add `saveOnClose` option to `openPrompt` - this will save any changes done in the palette when a user either clicks the palette icon again or when they change the cursor position in the document.
* Set the link palette to saveOnClose, also set it to close when clicking the palette icon when it's open.

## Related Issue

#45 

## Release notes
1. Adds image alt-text support.
2. Contextually available when image is selected.

**Demo:** https://ccc-alt-text--da-live--adobe.hlx.page/edit#/chrischrischris/da-milo-college/demo

Screenshot 2024-04-23 at 6 52 27 PM


auniverseaway (approved):
I think this works for now. My one small nit is that I think the icon should be 100% hidden if it cannot be used. The toolbar is going to get bigger over time and this would help cut down on size to make icons more contextual to the available tasks.

I used CSS `:has()` to hide assets. It's not 100% ideal, but it should be fine for now.

```
.ProseMirror-menuitem:has(.open-assets.ProseMirror-menu-disabled) {
  display: none;
}
```

auniverseaway (approved):

Issue #108, opened 2024-04-17
Provide UI to work with versions and audit entries


bosschaert:
We need the UI to work with Versions https://github.com/adobe/da-admin/issues/6 and Audit entries https://github.com/adobe/da-admin/issues/5

auniverseaway:
![Artboard 1](https://github.com/adobe/da-live/assets/1972095/08823edd-fa7e-4109-a2ad-767ef36ec4d1)

bosschaert:
I've updated the `versions` branch to look like the artboard:
<img width="1527" alt="Screenshot 2024-04-24 at 12 34 35" src="https://github.com/adobe/da-live/assets/226514/adc436a3-55c8-4c10-a0b9-34ea536f23db">

I couldn't find the icons on https://spectrum.adobe.com/page/icons/ so I left the old ones for now...  
Also note that there are no named versions yet.

Two questions @auniverseaway :
**How would we expand the list of audit logs at the top that are overflowing and hidden**?
<img width="308" alt="Screenshot 2024-04-24 at 12 37 06" src="https://github.com/adobe/da-live/assets/226514/7bd5abf3-7eae-4922-86b4-0ee06f832d24">

Just reveal them all when clicking on it?

**And how should we make visible the hidden audit log entries for previous dates**?
<img width="271" alt="Screenshot 2024-04-24 at 12 37 10" src="https://github.com/adobe/da-live/assets/226514/6df5542c-2788-4982-bb2a-e9909a52be5c">

Similarly reveal them all when clicking on it?

bosschaert:
I made a first attempt at naming versions. You can see it show up in the version list here beside the green circle. Right now I have a button to name the currently selected version in the toolbar but you probably want to change that 😉 
<img width="564" alt="Screenshot 2024-04-26 at 14 12 09" src="https://github.com/adobe/da-live/assets/226514/f897ff33-3453-4c50-9cd0-e1a936888dec">

@auniverseaway What UI interaction do you have in mind to enter the name for the version?

(note that for this to work some pending changes to da-admin are needed)

bosschaert:
> > Just reveal them all when clicking on it?

> Yup, I think so.

With https://github.com/adobe/da-live/commit/6bd5447ed76b879490ba76ac9906626d310d7abf I added expansion when clicking on it like this:

Initial visual:
[...]

PR #129, opened 2024-05-15
GH-108 - Versions & Audits

auniverseaway:
* Provide history pane in editor
* History pane shows both audits and versions
* Audits are grouped by day
* Audits show time & associated users
* Versions show time, users, and optional label
* Versions and audits of the same day are separate by design
* Provide ability to expand each date to see time and users
* Provide ability to click version icon to see a preview
* Ability to restore a version
* Ability to create a new labeled version

Resolves: GH-108

## How Has This Been Tested?

Manual E2E (for now)

https://ver--da-live--adobe.aem.live/edit#/aemsites/da-block-collection/drafts/cmillar/test
https://ver--da-live--adobe.aem.live/edit#/aemsites/da-block-collection/drafts/cmillar/net-new-no-version-or-audit

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.


auniverseaway:
## Some notes & quirks

1. I'm not terribly excited about the hover on "history" to see you can close the history view, but I also don't want to have the previous gray menu because when we start showing editor + preview + versions, we'll have to make major UX changes.
2. The create version API does not provide the location of the version created, so you cannot make a 1:1 entry without re-hitting the full list API. This means if you click restore on a version you just created (not a happy path), it will not work. To fix: the version API should return the location of the version on create.
3. If you make no audit-able changes, but create a version, the previous version label will be wiped in favor of the new version. This can be confusing given that there is a new entry added to the list. If we can get item 2 solved, we can check against the existing list to see if that version got updated and act correctly. The alternative would be to just not overwrite the previous version... which may be less confusing overall. I make a version, I get a version... even if it's the same as the other one.
4. We also do not know who is associated with the version: the 201 should come back with this info as well.
5. Audit log seems un-predictable. Cannot tell if it is time based, character based, de-bounced with collab weirdly or what.

## Todo

1. Get version on preview built.
2. Unit tests.

bosschaert (approved):

karlpauls:
@auniverseaway, I agree about #2 - the 201 should return the url of the version in the Location header (/cc @bosschaert )

Regarding #3, I'm not sure I follow - if you version again, it will override the previous version so there shouldn't be a new entry in the list. I suppose we could make it a new version and just have it be duplicated content but that would be something purely to accommodate the ability to create versions with a label. Alternatively, we could fail if you attempt to create a version without any change to the previous one (would be a boolean change in da-admin). 

Regarding #4, I guess we could make that some meta-data in the response body of the 201.

Regarding #5, not sure what you mean exactly. The audit log captures all changes in da-admin. If those came via collab, that means they can have more than one author and yes, they will be debounced (collab does wait up to a couple of seconds before writing back). Is that what you mean?

bosschaert:
> @auniverseaway, I agree about 2. - the 201 should return the url of the version in the Location header (/cc @bosschaert )
> 
> Regarding 4., I guess we could make that some meta-data in the response body of the 201.

Created https://github.com/adobe/da-admin/issues/45 for this.

> Regarding 3., I'm not sure I follow - if you version again, it will override the previous version so there shouldn't be a new entry in the list. I suppose we could make it a new version and just have it be duplicated content but that would be something purely to accommodate the ability to create versions with a label. Alternatively, we could fail if you attempt to create a version without any change to the previous one (would be a boolean change in da-admin).

Created https://github.com/adobe/da-admin/issues/46 for this.

PR #184, opened 2024-06-13
Playwright test stability

bosschaert:
## Description

Stabilize the Playwright tests

## Motivation and Context

Especially in the context of GH actions the PW tests are sometimes flaky.

The GH actions are run in a single core context which means that the browser runtimes run by PW have less cycles to catch up with any changes that are happening asynchronously. To give the browsers a little more time a few extra wait moments have been added to the tests.

## How Has This Been Tested?

These are tests

## Types of changes

## Checklist:

### DIFF
--- a/test/e2e/tests/copy_rename.spec.js
+++ b/test/e2e/tests/copy_rename.spec.js
@@ -27,23 +27,23 @@ test('Copy and Rename with Versioned document', async ({ page }, workerInfo) => 
   await page.locator('div.ProseMirror').fill('First text');
 
   // Wait 3 secs to ensure its saved in da-admin
-  await page.waitForTimeout(3000);
+  await page.waitForTimeout(5000);
 
   // Add some more text
   await page.locator('div.ProseMirror').fill('Versioned text');
-  await page.waitForTimeout(3000);
+  await page.waitForTimeout(5000);
 
   // Create a new stored version called 'myver'
   await page.getByRole('button', { name: 'Versions' }).click();
   await page.locator('button.da-version-btn').click();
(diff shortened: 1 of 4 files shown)


test/e2e/tests/copy_rename.spec.js
```diff
@@ -27,23 +27,23 @@ test('Copy and Rename with Versioned document', async ({ page }, workerInfo) =>
   await page.locator('div.ProseMirror').fill('First text');
 
   // Wait 3 secs to ensure its saved in da-admin
```
bstopp, line 29: Comments don't match the time. Is there a way to wait for an event to complete, rather than setting a time out?

bstopp:
Side note: i think `.spec.` is a Ruby convention? `.test.` is Node's convetion

test/e2e/tests/copy_rename.spec.js
bosschaert: Good catch on the comment - fixed.

On the Playwright level I haven't found another solution, although you can wait for things to appear, this takes cycles as well. I think that the problem here is that the PW browsers are running in the same GH action environment that just has very low CPU capacity, and I only found that giving the system some breathing space like this works consistently. 
But - if there is a better solution I would be all for it 😄

bosschaert:
> Side note: i think `.spec.` is a Ruby convention? `.test.` is Node's convetion

Yes right, and our unit tests use .test. but the Playwright convention seems to be to use .spec. so that's what I followed, see here https://playwright.dev/docs/writing-tests and a .js specific use-case https://www.lambdatest.com/learning-hub/playwright-javascript

bstopp (approved):

PR #296, opened 2024-12-14
Sort slash results so best match is first

chrischrischris:
Closest match that starts with slash input first, then closest match within string then the rest alpha sorted


auniverseaway (approved):
This is _so_ unnecessary, but 👍.

PR #372, opened 2025-03-11
Slash menu table edit commands

hannessolo:
The slash menu allows editing blocks (remove row, add row, etc).

Still missing:

- [x] Merge cells
- [x] Styling to match new sidebar icons

https://github.com/user-attachments/assets/4aedeac5-be09-4e7a-a818-f4216d92eef8

## Release notes

* Adds edit table items to slash menu.

Screenshot 2025-03-14 at 9 31 56 AM

### DIFF
--- a/blocks/edit/prose/plugins/slashMenu/slash-menu.css
+++ b/blocks/edit/prose/plugins/slashMenu/slash-menu.css
@@ -90,11 +90,11 @@
 }
 
 .slash-menu-icon.insert-table {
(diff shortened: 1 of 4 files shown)


chrischrischris (commented):
Nice!  Should we have a small arrow / caret showing there's a submenu?

Also would be nice if it could filter to the table commands.

auniverseaway (approved):

PR #426, opened 2025-04-23
[ignore] testing branch short name

kptdobe:
Test branch: https://rshort--da-live--adobe.aem.live/


kptdobe:
ok, so short name branches are really a requirement for the e2e tests...

Issue #421, opened 2025-04-17
[permissions] check for email addresses should be caseless


davidnuescheler:
i think it would be great if email addresses would be compared caseless (or lowercased) as a lot of customers give us email addresses that include arbitrary uppercase characters.

PR #428, opened 2025-04-24
chore : normalize email addresses before checking

kmurugulla:
## Description

- Added utilities to convert email address to lowercase , comparison

## Motivation and Context

#421 

## How Has This Been Tested?

- created test case for email comparison

## Types of changes

## Checklist:


auniverseaway (approved):
**Two things:**

1. This is great.
2. Nit: it's not needed right now.

The original ticket was about permissions which would be a `da-admin` issue.

I think throwing this in utils is probably a good idea, but we're also a bit sensitive to anything that could degrade editor performance. This PR introduces a new import inside `prose/index.js` which means another chain and more bytes over the wire before we get to editing.

Our editor is _by far_ our slowest in CVW. I think we still merge this... the hope would be that the file is already cached from being used elsewhere, but it's good knowledge for you to have that performance must be thought of in anything we do. Even at the expense of being DRY.

auniverseaway (changes requested):
Re-reading this I think we want to close this for now.

I didn't realize we already had a `toLowerCase` in both `prose/index.js` and `da-list-item.js`. This creates more code to maintain and a new import for the editor.

There's also an exported function that's only purpose is to satisfy a test.

My vote would be that we do not need this extra complexity at this time given that the original issue is within da-admin.

If we get to a point where we need to return an empty string, we can re-evaluate.

blocks/edit/prose/index.js
```diff
@@ -205,7 +206,7 @@ export default function initProse({ path, permissions }) {
         wsProvider.awareness.setLocalStateField(
           'user',
           {
-            color: generateColor(profile.email || profile.userId),
+            color: generateColor(normalizeEmail(profile.email) || profile.userId),
```
auniverseaway, line 209: Misread. This **_is_** missing `toLowerCase`. I would simply add it here and be done. No need to make a utility for a built in browser function.

auniverseaway:
Stale

Issue #422, opened 2025-04-21
[assets] Granular org fallback


auniverseaway:
As of right now, the org fallback only works if you have no site config. We should inherit on a row basis so that if a site has a single property (`aem.assets.link.type`) the rest still falls back to org so you don't have to duplicate your repoId, etc.

Work is in da-assets.js. Good first issue.

sourabratabose:
I am eager to contribute to this issue. Can I get assigned to this ?

sourabratabose:
I have submitted a PR which resolves this issue. Please review it and provide feedback if any.

Looking forward to the PR review.

PR #429, opened 2025-04-24
Updated fetchValue() function to handle repo level configuration fall…

sourabratabose:
…back to org level configuration as fetched

## Description

I have updated the fetchValue() function body to fetch both the organization level and repo level configurations which are then merged and returned on call. Repo level configurations as found will be considered as final over organization level configurations.

## Related Issue

https://github.com/adobe/da-live/issues/422

 

## Motivation and Context

As of right now, the org fallback only works if you have no site config. We should inherit on a row basis so that if a site has a single property (aem.assets.link.type) the rest still falls back to org so you don't have to duplicate your repoId, etc.

To solve this as mention in issue #422.

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x ] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.

### DIFF
--- a/blocks/edit/da-assets/da-assets.js
+++ b/blocks/edit/da-assets/da-assets.js
@@ -13,10 +13,6 @@ const ASSET_SELECTOR_URL =
 
 const CONFS = {};
 
-async function getMergedConf(owner, repo) {
-  return mergedConf;
-}
-
 async function fetchConf(path) {
   if (CONFS[path]) return CONFS[path];
   const resp = await daFetch(`${DA_ORIGIN}/config${path}`);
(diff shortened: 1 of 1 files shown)


blocks/edit/da-assets/da-assets.js
```diff
@@ -8,10 +8,15 @@ const { loadStyle } = await import(`${getNx()}/scripts/nexter.js`);
...
 
-const ASSET_SELECTOR_URL = 'https://experience.adobe.com/solutions/CQ-assets-selectors/assets/resources/assets-selectors.js';
+const ASSET_SELECTOR_URL =
+  'https://experience.adobe.com/solutions/CQ-assets-selectors/assets/resources/assets-selectors.js';
 
 const CONFS = {};
 
+async function getMergedConf(owner, repo) {
+  return mergedConf;
+}
```
chrischrischris, line 18: This doesn't appear to be used and `mergedConf` is undefined?

blocks/edit/da-assets/da-assets.js
```diff
@@ -27,7 +32,29 @@ async function fetchConf(path) {
 async function fetchValue(path, key) {
   if (CONFS[path]?.[key]) return CONFS[path][key];
```
chrischrischris, line 33: The code now ignores the `path` value passed in and only pulls the paths defined on lines 36 & 37.   I see that is only called once with the `/${owner}/${repo}/` path. 

If fetchValue is only ever called with that path, then do we even need to pass the path in?

blocks/edit/da-assets/da-assets.js
```diff
@@ -27,7 +32,29 @@ async function fetchConf(path) {
...
+  if (repoConf) {
+    if (data === null) data = {};
+    repoConf.forEach((item) => {
+      data[item.key] = item.value;
+    });
+  }
+
   if (!data) return null;
 
   const confKey = data.find((conf) => conf.key === key);
```
chrischrischris, line 60: `data` was an array previously, and is now an object which means that the `data.find` will fail on line 60.  Now that it's an object, you can just do a lookup

```suggestion
  const data = {};

  // Start with organization-level config
  orgConf?.forEach((item) => {
    data[item.key] = item.value;
  });

  // Override with repository-level config
  repoConf?.forEach((item) => {
    data[item.key] = item.value;
  });

  return data[key];
```

blocks/edit/da-assets/da-assets.js
```diff
@@ -39,23 +66,29 @@ async function fetchValue(path, key) {
...
 
   // Determine publicly available asset origin
-  const prodOrigin = await getConfKey(owner, repo, 'aem.assets.prod.origin') || `${repoId.replace('author', 'publish')}`;
+  const prodOrigin =
+    (await getConfKey(owner, repo, 'aem.assets.prod.origin')) ||
+    `${repoId.replace('author', 'publish')}`;
 
   // Determine if images should be links
-  const injectLink = (await getConfKey(owner, repo, 'aem.assets.image.type')) === 'link';
+  const injectLink =
```
chrischrischris, line 90: This change and below don't need to be refactored

chrischrischris (changes requested):

blocks/edit/da-assets/da-assets.js
sourabratabose: Sorry, I originally authored this function to abstract the merging of configurations away from fetchValue() function. I forgot to remove the function declaration after moving the body of the function inside fetchValue() itself.

blocks/edit/da-assets/da-assets.js
sourabratabose: Sorry for the mistake I made. I was so engrossed in the logic of merging configurations I forgot to account for the path.

blocks/edit/da-assets/da-assets.js
sourabratabose: I will be updating the PR to treat data as an array to retain the expected data type of `data` variable. Pushing objects to an array and updating values inside them if the keys match afterwards, is my approach.

sourabratabose:
Hi @chrischrischris , I didn't really understand as to why the path variable is required to be passed anyways.

Since configuration merge is fetched from an exhaustive list of paths. I decided to check whether paths are matching with the expected org and repo level paths, fetch the configurations and merge as required.

All other fetchValue() calls which dont match that path patterns will still fall through and return the value as fetched in the original function body.

auniverseaway:
Closing. A) there's now conflicts, B) I think there was some missing context of the needs for this PR and I think we can accomplish the need in a simpler way.

Issue #378, opened 2025-03-20
Deleting a document while having an editor open on it recreates it empty


bosschaert:
When I have an editor open on a document and then in a different browse window delete that document, the document gets cleared out in the editor, but then it still gets recreated after a few seconds in storage (da-admin) with empty content.

We should not recreate the document in storage while the document is still empty. In other words if we have an editor tab open somewhere and we delete that document it should not automatically get recreated in storage. This is especially important when deleting a whole tree. At the end of the delete operation the tree should be gone and not still be there with some empty docs.

kptdobe:
@bosschaert Any update on this one ? Having worked a little bit with DA, this is pretty "critical", especially as tabs keep being opened, I quickly get one tab left opened and, even after deleting the full parent folder structure, the pages keep coming back...

kunwarsaluja:
@bosschaert  similar behavior (but the opposite) is also reported by customer/partner where if we try to create a folder, on refresh the newly created folder isn'

![Image](https://github.com/user-attachments/assets/8f83c388-b522-4a35-a2d1-54304c603b4c)

t available in the DA UI. It took atleast 10 tried to get the folder created finally.

dkuntze:
@bosschaert any idea what triggered the rollback?

kptdobe:
Re-opened as fix has been reverted.

PR #450, opened 2025-05-16
Use document GUID as prosemirror reference to avoid updating deleted documents

bosschaert:
## Description

Store the current prosemirror document in the YDoc under `prosemirror-${guid}` instead of the previous `prosemirror` key.
A new prosemirror document gets a new GUID, even if it was stored in the same location as where a previous document used to live. This means that if there was an old editor open somewhere, that is editing a document that has since been deleted, these edits are then ignored.

The property `prosemirror-guids` contains an array of GUIDs with timestamps. Only the newest timestamp is active, any older guids on that array are not active any more. This array is often scrubbed, do deleted documents on the same location often don't appear in the `prosemirror-guids` at all any more.

When creating a fresh new document in the editor, create a new GUID for it, using `crypto.randomUUID()`
When editing an existing document, the GUID of that document is returned from the HEAD request in the `x-da-id` header. So we can set that to the `prosemirror-${guid}` in that case.

Requires: https://github.com/adobe/da-admin/pull/135 and https://github.com/adobe/da-collab/pull/69
Also requires existing open browser DA editors to be refreshed.

## Related Issue

Fixes: #378

## How Has This Been Tested?

Locally with da-collab and da-live

## Types of changes

## Checklist:

### DIFF
--- a/blocks/edit/edit.js
+++ b/blocks/edit/edit.js
@@ -4,7 +4,7 @@ import initProse from './prose/index.js';
 let proseEl;
 let wsProvider;
 
-async function setUI(el, utils) {
+async function setUI(el, utils, guid) {
   const details = getPathDetails();
   if (!details) {
     // el.classList.add('no-url');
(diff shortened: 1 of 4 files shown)


bosschaert:
The playwright tests will fail until the dependent PRs are merged.

auniverseaway (approved):

PR #512, opened 2025-08-08
Add preview link to library block and template Menus (#497)

auniverseaway:
Contributing branch on @JasonHowellSlavin's behalf so we can test against prod content.

* Adds a preview link to both block and template sections of the library. Modifies the preview SVG to be in line with other SVGs used in da-library.js

* Adding space


auniverseaway:
@JasonHowellSlavin feedback:

1. Preview button should be elevated to the main block entry (next to the caret). I would make it a lighter gray than the caret and then correctly blue outline it on hover.
2. Preview should be aware of common library patterns and should correctly transpile to the AEM URL.
    1. content.da.live/{org}/{site}/{path} - Most common as we own this domain and it doesn't have CORs errors.
    2. admin.dalive/source/{org}/{site}/{path}.html - I've never seen this, but it's technically possible.
    3. main--{site}--{org}.aem.page/path - People who open their CORs can do this.
3. Link should open into a "fullsize-dialog" not a new window.

blocks/edit/da-library/helpers/helpers.js
```diff
@@ -232,3 +232,32 @@ export function delay(ms) {
   // eslint-disable-next-line no-promise-executor-return
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export const getMetadata = (el) => [...el.childNodes].reduce((rdx, row) => {
```
chrischrischris: This function can be imported from here (double check to verify it's exactly the same):
https://github.com/adobe/da-live/blob/main/blocks/edit/utils/helpers.js#L420

blocks/edit/da-library/da-library.js
```diff
@@ -287,6 +294,22 @@ class DaLibrary extends LitElement {
...
+  }
+
+  handlePreviewClose() {
+    this._blockPreviewPath = '';
+    this._previewItemName = '';
+  }
+
+  handlePreviewLoad() {
+    this.shadowRoot.querySelector('.da-fs-dialog-plugin').showModal();
+    this.shadowRoot.querySelector('.da-fs-dialog-plugin').classList.remove('hide');
```
chrischrischris: nit run selector once
```suggestion
    const fsDialogPlugin = this.shadowRoot.querySelector('.da-fs-dialog-plugin');
    fsDialogPlugin.showModal();
    fsDialogPlugin.classList.remove('hide');
```

PR #518, opened 2025-08-14
chore: fix test and lint

kptdobe:
Go back to a green state...


bosschaert:
@kptdobe the IMS login of the playwright tests fails. Could it be because the branch name `test-and-lint` is more than the max? IIRC the max branch name length was 8 characters.

PR #532, opened 2025-08-19
Skip auth tests when branch name is too long

chrischrischris:
Silly IMS rules

This will allow for running all non-auth tests even if the branch name is greater > 8 chars.


kptdobe:
I am not convinced about that... Goals of the tests is to make sure you did not break anything. Here, you would easily be able to break everything auth related (which is probably the worst case scenario).
If the intent is to still see the other tests running, even if you used a too long branch name, then I would rather move the auth tests at the end, so that the other tests run. But you would still be required to rename your branch before being able to merge to main.

auniverseaway:
I think we probably should revisit our IMS "don't call it regex" regex.

auniverseaway:
@chrischrischris I think I'm with @kptdobe on this... we can revisit our IMS regex rules to be more permissable and start with that. Less code we have to maintain in the long run and more likely to catch things that may break auth.

chrischrischris:
Yep agree fixing it in IMS is the way to go.

Issue #542, opened 2025-08-22
[utils] Support Nexter forks


auniverseaway:
As a developer, I would like the ability to supply a forked nx query param so that I don't have to have my branch live on the `adobe` github org to get feedback on my PR.

PR #543, opened 2025-08-22
GH-542 - Support Nexter Forks

auniverseaway:
* Ability to support nexter forks (`?nx={BRANCH}--{SITE}--{ORG}`)

Resolves: GH-542


auniverseaway:
This came up when working with @kmurugulla on the Media Library and with @rgclayton when working on the AEM Permissions app.

auniverseaway:
Given @andreituicu comments, which I now remember, I'll close this. We will have to merge in branches we believe to be of good faith into this fork. I can't think of another way that also guarantees the same level of security.

PR #544, opened 2025-08-27
feat: only load preview frame when needed

kptdobe:
Today, the editor keeps updating the preview panel even if it is not visible. A lot of unnecessary iframe dom updates, resource fetching and javascript execution is happening on content change even if the preview panel is not visible.

This moves the preview loading, the preview poll and the preview body update to when the preview panel is visible... 

Test: https://preview--da-live--adobe.aem.live/


chrischrischris (approved):
Nice - this was on my todo list.

Issue #545, opened 2025-08-28
Prevent duplicate resource creation on repeated clicks of action button


ajayullal:
I noticed an issue with the new resource creation flow. When I select a resource type (like folder or media) and click the action button multiple times (e.g., create folder or upload), it creates multiple instances of the resource. A potential fix could be to disable the action button until the creation process is complete.

![Image](https://github.com/user-attachments/assets/d2e30dbc-5f57-42f6-a713-029012159539)

PR #555, opened 2025-09-01
fix: disable action buttons during resource creation to prevent resource duplication

ajullal:
## Description
I noticed an issue with the new resource creation flow. When I select a resource type (like folder or media) and click the action button multiple times (e.g., create folder or upload), it creates multiple instances of the resource. A potential fix could be to disable the action button until the creation process is complete.

## Related Issue
https://github.com/adobe/da-live/issues/545

## How Has This Been Tested? (Screenshots)
Before: https://main--da-live--ajullal.aem.page/
After: https://disable-in-progress-buttons--da-live--ajullal.aem.page/

Before:
Bug

After:
fix

### DIFF
--- a/blocks/browse/da-new/da-new.css
+++ b/blocks/browse/da-new/da-new.css
@@ -51,7 +51,8 @@ button {
   transition: outline-offset .2s;
 }
 
-.da-actions-new-button:disabled {
+.da-actions-new-button:disabled,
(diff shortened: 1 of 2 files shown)


auniverseaway:
This is more code than we currently need for the issue. We'll pull this in house and address. Thanks for raising.

Issue #566, opened 2025-09-08
Improve block library tooltips


auniverseaway:
* Make touch / mobile friendly (no hover, sizing, spacing)
* Fit existing UX a little better
* Reduce code complexity

PR #567, opened 2025-09-08
Improve library tooltips

auniverseaway:
## Example
https://libtool--da-live--adobe.hlx.live/edit#/audemars-piguet/arbres-fondationsaudemarspiguet/en

## Release notes

* Make touch friendly - click instead of hover
* Make touch friendly - more spacing from add icon
* Make more consistent with existing UX.
* Toggle-able on & off
* Simplify CSS & JS
* Fix icon showing black on mouse out

Resolves: #566

Screenshot 2025-09-07 at 7 29 47 PM

### DIFF
--- a/blocks/edit/da-library/da-library.css
+++ b/blocks/edit/da-library/da-library.css
@@ -364,24 +364,48 @@ ul {
   transition: transform .2s ease-in-out;
 }
 
-.da-library-type-group-detail-item button {
-  display: flex;
-  justify-content: space-between;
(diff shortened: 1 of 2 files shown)


mhaack (approved):
I like it. Wonder if we should have an (query/config) option for the library to always show the extra description. Projects might prefer this. Hm, but this can be added if someone requests it.

asthabh23 (approved):
Thanks Chris, this looks nice!

PR #579, opened 2025-09-15
feat: log out user if not authorized

### LINKED ISSUE #563: Disconnect user if a 401 happens in collab
When a non-authorized user opens a page, it is forced to sign-in. But if while editing and for some unidentified yet reason (token expired, worker reloaded... ?) user is not authorized anymore, they can continue authoring for a long time. On page reload, all changes would be gone. 

The websocket could detect that case and force a re-signing.

We need to be careful because we observe a lot of `401` in the logs, I suspect the auth to not be fully stable. It might end up that users get disconnected way too frequently (which would be expected but a consequence of something wrong...)

See also thread https://cq-dev.slack.com/archives/C08MJQ0Q3GA/p1757005611676039

kptdobe:
Fix https://github.com/adobe/da-live/issues/563

Session might expire or user can be logged out for some reason but still continue to author. Editor does not detect this case and still allow user to edit for an infinite time (until page reload).

Test: https://kickout--da-live--adobe.aem.live/

- Sign in to a project
- Open a page for authoring
- Open a new tab with the explorer and sign out
- Go back to editor, make edits. Reload will happen after one or 2s and force user to sign-in again. Minimal changes are lost.

### DIFF
--- a/blocks/edit/edit.js
+++ b/blocks/edit/edit.js
@@ -36,7 +36,7 @@ async function setUI(el, utils) {
   }
 
   const { daFetch } = await utils;
-  const { permissions } = await daFetch(details.sourceUrl, { method: 'HEAD' });
+  const { permissions = [] } = await daFetch(details.sourceUrl, { method: 'HEAD' });
   daTitle.permissions = permissions;
   daContent.permissions = permissions;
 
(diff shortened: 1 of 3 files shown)


chrischrischris:
This forces the user into the ims signin screen, but is a bit confusing as it does it with no warning.  I wonder if we should have a dialog alerting the user with the option to signin again or go to da home?  @auniverseaway thoughts?

chrischrischris (approved):

auniverseaway:
Stale. Notable: I've seen an increase of un-expected logouts from DA / IMS recently.

PR #588, opened 2025-10-02
feat: use media route

kptdobe:
Use the new `/media` route - see https://github.com/adobe/da-admin/pull/178

Note: only merge this one when https://github.com/adobe/da-admin/pull/178 has been deployed to production.

Test: https://media--da-live--adobe.aem.live/?da-admin=stage

UX in case of error (like if image is > 20MB) needs to be reviewed.


auniverseaway:
I'd love to get this shipped somehow. Aside from the merge conflict, I think this only needs a few graceful fallbacks to cut down on support issues:

**Motivation: this change will have the following impact to the existing process:**
1. You cannot upload bigger than 20MB.
2. You cannot upload unless you're in a legitimate Edge Delivery site.
    1. no uploads to org level.
    2. no uploads _before_ Edge Delivery site made.
3. You cannot upload unless you have `basic_author` permissions for Edge Delivery.

**How I'd solve (happy to take other suggestions)**
For `no images above 20MB` - throw the modal. They can't get the image out the door anyway.

For `org-level uploads` - throw the modal. Let's see who it impacts.

For `no site === no upload` or `no basic_author === no upload` - graceful fallbacks:
1. Try to upload `/{org}/{site}/media/{original-drop-parent-path}/{something-unique}.png`.
2. Fallback to original behavior + sanitized name if the author cannot access `/{org}/{site}/media` for whatever reason.

Issue #586, opened 2025-09-29
Broken image icon while a dropped image is uploading


kptdobe:
Just drop a large image (slow to upload), you will see a broken image icon.
Check network tab, image points to: https://content.da.live/auniverseaway/da/assets/fpo.svg

@auniverseaway: SVG should be uploaded to this repo, no ?

PR #589, opened 2025-10-02
fix: repare uploading icon

kptdobe:
Fix https://github.com/adobe/da-live/issues/586

Test: https://fpo--da-live--adobe.aem.live/


auniverseaway (approved):
Sheesh. That's bad.

PR #632, opened 2025-11-10
Add DA metadata support

chrischrischris:
This adds support for document metadata that is only visible to DA.  When helix ingests the DA html, the metadata is ignored due to it being stored outside of `body > main`.

The metadata can be accessed via `getDaMetadata` and `setDaMetadata` in `utils/helpers.js`.

The data is stored under `body > div.da-metadata` as a key / value stored in the same format as helix tables:

```
<body>
  <main>...</main>
  <div class='da-metadata'>
    <div>
      <div>Key1</div>
      <div>Value 1</div>
    </div>
    <div>
      <div>Key2</div>
      <div>Value 2</div>
    </div>
  </div>
</body>
```

Using this metadata da-diff accept/reject choices are stored in the metadata under the keys: `acceptedHashes` and `rejectedHashes`

https://github.com/adobe/da-nx/pull/117 and https://github.com/adobe/da-collab/pull/104 can be merged after this PR is merged

### DIFF
--- a/blocks/edit/da-assets/da-assets.js
+++ b/blocks/edit/da-assets/da-assets.js
@@ -240,7 +240,9 @@ export async function openAssets() {
 
             const insertTypeSelection = cropSelectorWrapper.querySelector('.da-dialog-asset-structure-select input:checked');
             const structureConfig = !insertTypeSelection || insertTypeSelection.value === 'single' ? null : JSON.parse(decodeURIComponent(insertTypeSelection.value));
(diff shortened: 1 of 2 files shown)


mhaack:
@chrischrischris @auniverseaway is the plan to display "da-metadata" somehow in the editor?

 Is my reading correct, this is for da only metadata, not intended to be published to Helix? I think for DA & UE I can make good use to that as well.

PR #634, opened 2025-11-11
feat: provide word style list creation

shsteimer:
## Description

- If the user types "* " or "- " at the start of a line, create a UL
- if they type "1." or "1) " create an OL

This allows more of the muscle memory users have from word to be re-used in DA.

full disclosure, this was largely vibecoded, but looking over the code it makes sense to me and works when tested.

## Motivation and Context

Because I get annoyed when authoring that this doesn't work :)

## How Has This Been Tested?

tested locally in the editor at http://localhost:3000/edit#/aem-sandbox/block-collection/drafts/shsteimer/testing

contemplated unit tests, but no other key handlers seemed to have them, assume there is a reason for that.

## Types of changes

## Checklist:


blocks/edit/prose/plugins/keyHandlers.js
```diff
@@ -47,6 +48,34 @@ export function getURLInputRule() {
   );
 }
 
+/**
+ * Converts "* " or "- " at the start of a line into a bullet list.
+ *
+ * @param {Schema} schema - The ProseMirror schema
+ * @returns {InputRule|null} the bullet list input rule, or null if
+ *   bullet_list node doesn't exist
+ */
```
chrischrischris: nit - can simplify comment to just 
`// Converts "* " or "- " at the start of a line into a bullet list.`

same for `getOrderedListInputRule`

We don't minify the js files, sotry to keep size down wherever possible.

blocks/edit/prose/plugins/keyHandlers.js
```diff
@@ -82,6 +111,12 @@ export function getURLInputRulesPlugin() {
   return inputRules({ rules: [getURLInputRule()] });
 }
 
+// Returns an inputRules plugin for list formatting (bullet and ordered lists)
+export function getListInputRulesPlugin(schema) {
+  const rules = [getBulletListInputRule(schema), getOrderedListInputRule(schema)].filter(Boolean);
```
chrischrischris: I don't think there's ever a time where `schema.nodes` will NOT have `bullet_list` and `ordered_list`, so it should be safe to assume that they will be present.

I think that all of the listInputRule code can be consolidated into this function:
```
export function getListInputRulesPlugin(schema) {
  const rules = [];
  const olSchema = schema.nodes['ordered_list'];
  if (olSchema) rules.push(olSchema);
  const blSchema = schema.nodes['bullet_list'];
  if (blSchema) rules.push(blSchema); 
  return inputRules({ rules });
}
```

shsteimer: makes sense to me, but if `schema.nodes.bullet_list` and `schema.nodes.ordered_list` will always be defined, then we don't need the `if` statements at all right?

Assuming that's true, I simplified it to just:

```
export function getListInputRulesPlugin(schema) {
  const rules = [
    wrappingInputRule(/^\s*1[.)]\s$/, schema.nodes.ordered_list),
    wrappingInputRule(/^\s*([-*])\s$/, schema.nodes.bullet_list),
  ];
  return inputRules({ rules });
}
```

If you think the ifs are needed, let me know and happy to update to the suggestion your provided.

PR #644, opened 2025-11-17
Navigation Enhancements: Cache large folders and show loading message when slow

chrischrischris:
1. When loading a new folder takes over 1s, shows a "Loading..." message on the screen
2.  When there are more than 40 items in a folder, the listing will be cached in sessionstorage.  This makes it much faster when re-navigating to that folder.  If there have been changes to the folder, it will update once the fetch comes back.

### DIFF
--- a/blocks/sheet/da-sheet-tabs.js
+++ b/blocks/sheet/da-sheet-tabs.js
@@ -123,7 +123,7 @@ class DaSheetTabs extends LitElement {
       return;
     }
     if (e.submitter.value === 'confirm') {
-      const name = Object.fromEntries(new FormData(e.target))?.name;
+      const name = Object.fromEntries(new FormData(e.target))?.name?.trim();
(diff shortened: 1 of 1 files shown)


auniverseaway (changes requested):
I would for sure remove the caching piece. This feature alone would generate a very large amount of support tickets due to confusion. I also don't think it buys us much when the majority of lists take less than 750ms. I'm impatient, but even I can deal with 750ms...

Slow lists are like 0.001% (probably less TBH) of all content, so we are adding a bunch of code here we will have to maintain for something that is extremely uncommon. To the point where I have only had 3 complaints about slow lists and all three were about da-bacom.

I also wouldn't merge this until we know S3 will be as bad for these larger lists. We are moving from R2 to S3, and if this is on the horizon (it is), we would likely never come back and remove all this code.

PR #660, opened 2025-11-21
Floating toolbar implemented

ravuthu:
Showing a toolbar when selecting the copy text might be convenient for authors to edit content

## Motivation and Context

I was authoring a block that had around 20 cards, and each card had an H3. Clicking on the edit link was an additional effort that I had to put in every time I format H3. Having a floating toolbar would solve that, IMHO.

## How Has This Been Tested?

Initialized the localhost and tested it on http://localhost:3000/edit#/aem-sandbox/block-collection/aclarke. Made sure that it will be hidden when selecting an option or clicking outside the toolbar, etc. It will not be shown when the user selects a link or an image. I verified the feature on mobile also.

Screenshot 2025-11-21 at 9 37 12 AM

[Edit aclarke - DA.webm](https://github.com/user-attachments/assets/b20c2650-78cd-43a8-b2cc-f36f88c6760b)

## Types of changes

## Checklist:

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -10,6 +10,7 @@ module.exports = {
   },
   rules: {
     'import/no-unresolved': 'off',
+    'import/no-cycle': 'off',
     'no-param-reassign': [2, { props: false }],
     'linebreak-style': ['error', 'unix'],
     'import/extensions': ['error', { js: 'always' }],
--- a/blocks/browse/da-new/da-new.js
+++ b/blocks/browse/da-new/da-new.js
@@ -111,8 +111,8 @@ export default class DaNew extends LitElement {
     const formData = new FormData(e.target);
     const split = this._fileLabel.split('.');
     const ext = split.pop();
-    const name = split.join('.').replaceAll(/\W+/g, '-').toLowerCase();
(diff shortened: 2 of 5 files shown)


auniverseaway:
I think this is really interesting. It makes me feel a bit uncomfortable, but I do think its great research.

I always think about, "The Windows 98 problem" which is... having many different ways (or UX patterns) to do the same task. It becomes a support nightmare and confusing for practitioners. If you look at the screenshot below, how many ways can you open up Internet Explorer? 3. That's three ways they had to document, write code around, etc.

Having **_another_** toolbar... in addition to the slash menu (which is already another UX), means we have another place to do all our menu updates and we have more documentation to write, etc. Again, same problem as Windows 98 and we're only a lowly editor.

I'm tempted to say, "if we ship this toolbar... can we remove it from the side menu?" I'm not sure if this is even a good idea. You're on an empty line and what do you do?

**_For now_**, I would opt to say, _"let's increase our keyboard shortcut footprint"_ over shipping another UX paradigm. If we see that this crops up as an issue for more users, we have this in our pocket.

I would also say I'm still not opposed to popping out the existing menu without the main menu labels. This would be a nice middle ground and solves some of the issues I mention above.

image

PR #668, opened 2025-11-25
Update based on table name changes

chrischrischris:
Also handle svg imgs explicitly


auniverseaway (approved):

auniverseaway (changes requested):
We shouldn't pin SVGs to 150px. I would opt for the solution we came up with yesterday on our call with @rgclayton.

Issue #669, opened 2025-11-25
[live preview] Set cursor context


auniverseaway:
Context: https://github.com/adobe/da-live/discussions/379

1. Fixed 4-5s delay of live preview
2. Gave live preview a fixed device height - resolving 100vh issues.
3. Set live preview scroll position based off cursor in document
4. Warn author that live preview could not start
5. Obfuscate when a document 404s because it doesn't exist

PR #670, opened 2025-11-25
GH-669 - Set live preview cursor context

auniverseaway:
### Main features
1. Fixed 4-5s delay of live preview
2. Gave live preview a fixed device height - resolving 100vh issues.
3. Set live preview scroll position based off cursor in document
4. Warn author that live preview could not start
5. Obfuscate when a document 404s because it doesn't exist

### Additional notes / context

* Removed glue for `startPreviewing` etc.
* Simplified da-content styles.
* Simplified da-content logic.
* Allow editor to collapse down to 375px if live preview needs it
* PostMessage initialization logic is more reliable
* Removed setHeight as this is no longer needed.
* Cache current doc in case live preview is opened (5s fix)
* dapreview.js now uses scrollIntoView which a modern scroll impl
* scrollIntoView requires setting fixed position temporarily to avoid parent scrolling (may need further tweaks on timing for low CPU devices)
* Only push content on cursor changing ProseMirror "block" elements
* Turn previous polling technique that powered live preview into a proper plugin to remove some jank.
* Fixed "mismatch transaction" warning when hash changes
* Introduced a new `getHtmlWithCursor` function in prose2aem to handle reliable cursor positioning.
* Minor additions to da-y-wrapper (NodeSelection)
* Handle scroll in dapreview.js (inside iframe)
* Send back message to parent when DOM has updated (previously only height was sent back)

1. Only change the cursor when it changes - currently still sending entire doc.
2. Bidirectional cursor position is not supported
3. Showing cursor in live preview
4. BYO live preview domain
5. Auto open live preview based on config

### Test
Before: https://da.live/edit#/aemsites/author-kit/index
After: https://livepre--da-live--adobe.aem.live/edit?ref=livepre#/aemsites/author-kit/index
After: https://livepre--da-live--adobe.aem.live/edit?ref=livepre#/aemsites/da-block-collection/index (graceful death)

Resolves: GH-669

## How Has This Been Tested?

E2E - Planning on adding Playwright over break...

### Test scenarios
1. Open live preview without interacting with document - shows live preview
2. Move cursor - live preview changes
3. Move cursor one character in same block element (para, list item) - live preview does not change
4. Scroll down document - live preview is pinned with an offset of 36px
5. Click into new block element - live preview changes
6. Allow scrolling inside live preview
7. Click block name (which converts to class name) - live preview updates
8. Switch device for live preview - still pinned in upper corner of preview
9. Updates should not cause parent page to scroll (which is a known issue with scrollIntoView)
10. Closing live preview re-engages document tabs (preview, versions, ue)
11. Warn when live preview cannot start
12. Versioning pane still opens and works as expected (maybe a little broad)
13. Switch document and live preview should work with new document

Screenshot 2025-11-25 at 10 38 33 AM

## Types of changes


blocks/edit/da-content/da-content.js
```diff
@@ -109,19 +77,31 @@ export default class DaContent extends LitElement {
           @versionreset=${this.handleReset}>
```
rgclayton: Does this need changed since you deleted handleReset()?

PR #672, opened 2025-11-25
fix: url poisoning via user-controlled input

rofe:
- Extended `sanitiseRef()` to `sanitizeName()` so it can handle file names and extensions
- Added utility functions `sanitizePath()` and `sanitizePathParts()`
- Updated places in the code where paths were being manipulated to use one of the sanitizing functions
- Added test coverage

https://j33256--da-live--adobe.aem.page/edit


rofe:
Thanks @chrischrischris. I think it’s related to my sanitizing paths…

PR #692, opened 2025-12-11
feat: delay versions loading

kptdobe:
Versions are loaded after document is opened. If no one clicks on the Version History button, this was useless. Number of clicks on the button is very low.

If a document has more than 500+ versions (including empty version files for edition tracking), `/versionlist` fires more than 500 requests (get object, 500x get versionsource...)! It is pretty common that a document has more than 100 or 200 versions, especially documents with active authors...

Delaying the version loading will save thousands of requests!!

I have added a nice "Loading..." experience.

Before: https://main--da-live--adobe.aem.live/edit#/auniverseaway/oh-demo/index
After: https://versions--da-live--adobe.aem.live/edit#/auniverseaway/oh-demo/index


auniverseaway (approved):
This works.

The likely better solution would be to not render the `da-versions` web component if its not the currently requested pane. Then you can avoid touching da-versions altogether and passing props. Pretty sure this was how it was done before I introduced this regression.

PR #707, opened 2025-12-18
fix: add org level permissions to new org

### LINKED ISSUE #415: [Onboarding] Make the user that creates the website the "admin" of it
# Context

Currently, when a new site is created in DA using https://da.live/start , it is created public by default.
Users need to take additional action to configure permissions in order to restrict who can read or edit the content.
This can create confusion for users, if they are not aware.

# Idea

When a new site is created, the user that does it is immediately made the "admin" of it.

# Details

As part of the website creation we can create a permissions config sheet with the user's IMS ID that they are logged in with and activate it.

| path | groups  | actions |
|------|--------|--------| 
| /+**  | ${user_id} |  write |
| CONFIG	| ${user_id} | write |

This would effectively make every new site's content private by default. 
If the user wants their content to be public and for everybody to be able to read and edit it, they can remove the permissions.

rofe:
Fix #415 

https://i415--da-live--adobe.aem.page/

### DIFF
--- a/blocks/start/start.js
+++ b/blocks/start/start.js
@@ -32,38 +32,78 @@ const AEM_TEMPLATES = [
   },
 ];
 
-const ORG_CONFIG = {
-  data: {
-    total: 2,
-    limit: 2,
-    offset: 0,
-    data: [{ key: '', value: '' }],
-    ':colWidths': [169, 169],
-  },
-  permissions: {
-    total: 2,
-    limit: 2,
-    offset: 0,
-    data: [
-      {
-        path: 'CONFIG',
-        groups: '$EMAIL$',
-        actions: 'write',
-        comments: 'The ability to set configurations for an org.',
-      },
-      {
-        path: '/ + **',
-        groups: '$EMAIL$',
-        actions: 'write',
-        comments: 'The ability to create content.',
-      },
+const ORG_CONFIG = `{
+    "data": {
+        "total": 1,
+        "limit": 1,
+        "offset": 0,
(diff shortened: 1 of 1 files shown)


blocks/start/start.js
```diff
@@ -125,9 +140,82 @@ class DaStart extends LitElement {
...
+        },
+        ':names': ['data', 'permissions'],
+        ':version': 3,
+        ':type': 'multi-sheet',
+      };
+
+      const body = new FormData([['config', JSON.stringify(configJson)]]);
+      const orgResp = await daFetch(orgUrl, { method: 'PUT', body });
+      if (!orgResp.ok) {
+        if (orgResp.status === 401 || orgResp.status === 403) {
```
kptdobe: Can this really happen ?

rofe: I thought it prudent to check for... unless we fully trust the `window.adobeIMS.isSignedInUser()` above?

blocks/start/start.js
```diff
@@ -140,9 +213,39 @@ class DaStart extends LitElement {
 
   async submitForm(e) {
     e.preventDefault();
+    const siteUrl = e.target.action;
+
+    // Check if user is signed in
+    if (!window.adobeIMS?.isSignedInUser()) {
```
auniverseaway: They cannot visit this page without being signed in...

Pull up https://da.live/start in an incognito browser, it will boot you to IMS.

auniverseaway (approved):
This is clever. I love this.

PR #727, opened 2026-01-22
fix mark order

### LINKED ISSUE #627: [edit] Enforce Helix order for link decoration in live preview
Helix has a specific order it creates tags: `<strong><em><a>` etc.

ProseMirror creates an arbitrary order based on when / how the author decorates their link. When content is pushed onto the page during live preview, this order is different than Helix. Projects that expect the Helix order will not show buttons as expected.

## Criteria of acceptance
1. Live preview link decoration order matches Helix (to be determined w/ Helix team)

hannessolo:
Fix #627 
Screenshot 2026-01-22 at 14 47 36

https://i627--da-live--adobe.aem.live

Screenshot 2026-01-22 at 16 08 20

### DIFF
--- a/blocks/edit/prose/schema.js
+++ b/blocks/edit/prose/schema.js
@@ -250,12 +250,6 @@ const baseNodes = {
 };
 
 const baseMarks = {
-    code: {
-    parseDOM: [{ tag: 'code' }],
-    toDOM() {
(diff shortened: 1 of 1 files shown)


hannessolo:
@mhaack this PR should also fix the button formatting in quick edit

PR #732, opened 2026-01-26
feat: show save button directly when editing configs

### LINKED ISSUE #704: [config] da-title web component should show save
`da-title` is shared across all our main editing views: `edit` `media` `sheet` `config`. In the case of config, we should not show the paper airplane and should instead show the save button. This will hopefully add clarity to admins who forget to "save" their config because they think its collaborative and should save automatically.

We explicitly require a save for config because the last thing you want is to save a config mid typing.

usman-khalid:
## Summary

- Refines `da-title` action behavior by context: hides actions for docs under `/.da/` and allows save only for configs
- Adds config staleness detection with polling and a modal prompt when remote config changes are detected during editing. User can choose to ignore or refresh

**Other changes:**
- Updates the buttons in the component to have a pointer cursor to indicate they're clickable

## Related Issue

Fixes #704 

## How Has This Been Tested?

Tested on locally and on https://i704--da-live--usman-khalid.aem.page

image
image
image

## Types of changes

## Checklist:

### DIFF
--- a/blocks/edit/da-library/helpers/helpers.js
+++ b/blocks/edit/da-library/helpers/helpers.js
@@ -278,6 +278,7 @@ export async function getPreviewStatus({ org, site, pathname }) {
     const json = await aemAdmin(path, 'status', 'GET');
     return json.preview.status === 200;
   } catch (err) {
+    // eslint-disable-next-line no-console
     console.log(`Could not get preview status for ${path}`, err);
     return null;
   }
--- a/blocks/edit/da-title/da-title.css
+++ b/blocks/edit/da-title/da-title.css
@@ -33,6 +33,7 @@ da-dialog {
   line-height: 16px;
   padding: 5px 14px;
   text-decoration: none;
+  cursor: pointer;
 }
 
 .con-button.blue {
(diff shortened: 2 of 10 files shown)


auniverseaway:
@usman-khalid thanks for the reminder. I disagree with myself from time to time. :-)

If you want to ship this, I would simply fill the existing blue button with the word "Save" instead of everything else. There's work to do on the cloud regarding current state, but it requires more consideration since we want to let authors know if something has changed while they have the sheet open. My loose idea is:

1. Upon open, cache the current sheet.
2. Poll every two minutes to see if sheet has changed.
3. If changed, warn the author.

We can super gold plate-y with the above:

1. Offer to refresh
2. Offer to compare
3. Offer to stash changes

I would avoid most of these and simply pop a modal and offer to refresh or ignore. I do still think we wait for @hannessolo's collab PR so we get a sense of how config is impacted.

auniverseaway (changes requested):
The user experience seems perfect. We need to move a few of the bits around, but otherwise looks really good.

usman-khalid:
Closing in favour of #877

PR #741, opened 2026-01-30
perf: split plugin loading to get editor rendered faster; GET  the doc for permissions and metadata check

chrischrischris:
1. Do async plugin loading
2. Change getting permissions from HEAD to GET in order to prep for BYO Editor.  Also preps for future perf enhancement of rendering doc in prosemirror before websocket connection finishes.
3. Add ESM Module loading error detection, which then sets a da-cache=off cookie to force the browser to not use the cache (pending cloudflare worker update).

After merging this PR, this nexter PR can be merged: https://github.com/adobe/da-nx/pull/160


head.html
```diff
@@ -14,4 +14,10 @@
 }
 </script>
 <script nonce="aem" src="/scripts/scripts.js" type="module"></script>
+<link rel="preconnect" href="https://auth.services.adobe.com">
+<link rel="preconnect" href="https://collab.da.live">
+<link rel="modulepreload" href="/deps/da-y-wrapper/dist/index.js">
+<link rel="modulepreload" href="/deps/lit/dist/index.js">
+<link rel="modulepreload" href="/deps/da-parser/dist/index.js">
```
auniverseaway: I'd remove all these. While this will make the editor faster, it will make everything else slower.

auniverseaway (changes requested):

auniverseaway (approved):
Awesome. Really looking forward to this.

Knowing IMS is coming up sooner in the call stack likely helps an edge case race condition we've seen.

PR #743, opened 2026-02-02
fix: wrong status call

kptdobe:
We see a lot of 404 going to urls like https://admin.hlx.page/status/source/doe-app-svc/main/da-global/docs/library/templates/transition-to-primary-school

Note the `/status/source` which is an invalid API call. The fix is removing the wrong check, the check is done in the `getEdsUrlVars`.

This is the "preview status check" introduced https://github.com/adobe/da-live/pull/512. Not sure why this is useful...

### DIFF
--- a/blocks/edit/da-library/da-library.js
+++ b/blocks/edit/da-library/da-library.js
@@ -544,10 +544,8 @@ class DaLibrary extends LitElement {
       try {
         const itemUrl = new URL(getUrl(item));
         path = itemUrl.pathname;
-        if (itemUrl.origin.includes('--')) {
-          const [org, site] = getEdsUrlVars(getUrl(item));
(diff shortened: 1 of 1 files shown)


auniverseaway:
Thanks for fixing this.

> Not sure why this is useful...

Super authors put DA URLs (content.da.live / admin.da.live) in their library configs. The bug came from a PR that introduced a "preview" button that pulls up an iframe to show the aem.page version of the page. If the page has never been previewed, the author would get a 404 because we convert the DA-flavored URL into an aem.page URL.

It's early, but I'll have to look at why these 404s were so noisy. We should only be doing the 404 check on click. Even if it means the preview modal pop up is a little slower or we have to show a loading screen.

PR #779, opened 2026-02-06
feat: sheets collab

hannessolo:
https://i701--da-live--adobe.aem.live/

## Release Notes
1. Enables real-time collaboration for sheets
2. Does not persist changes until an author moves off a cell

**Note:** DA Config will continue to be non-collaborative as persistence is handled differently for permissions.


blocks/sheet/collab/index.js
```diff
@@ -0,0 +1,76 @@
...
+        color: generateColor(profile.email || profile.userId),
+        name: profile.displayName,
+        id: profile.userId,
+      });
+    });
+  } else {
+    wsProvider.awareness.setLocalStateField('user', {
+      color: generateColor(`${wsProvider.awareness.clientID}`),
+      name: 'Anonymous',
+      id: `anonymous-${wsProvider.awareness.clientID}}`,
```
chrischrischris: Extra `}` at end?

chrischrischris (dismissed):
Awesome work!  We probably want to remove a lot of the `console.log`s but if we want to keep at first I'm good with that.  Just have a followup PR later.

auniverseaway (dismissed):

bosschaert:
I played a little bit with this, one client running in Chrome and the other in Firefox. I tested near-simultaneous edits on the single-sheet files and multi-sheet files in the same and different sub-sheets.
It worked all fine for me, no surprising behaviour and a nice collab experience. 

So, LGTM! 👍

PR #810, opened 2026-02-26
feat: keep selection when opening library

kptdobe:
Open a document for editing, select some content and click on the Library icon: the selection disappear.

Root cause: the focus is set on the search input of the library.

Browser cannot have a focused element and a selection at the same time. Gemini created this proposal which is to create a fake overlay. Not perfect and pretty complicated.

Test url: https://fakesel--da-live--adobe.aem.live/


auniverseaway (changes requested):
Small nit. Otherwise looks good. Glad someone is using this feature and polishing it down to be truly useful.

PR #835, opened 2026-03-09
feat: add page scheduler functionality to DA

amol-anand:
- Adds an optional Schedule Publish action in the Prepare menu with datetime selection and existing schedule visibility.
- Schedules page publish after registration + preview checks, with correct sheet/non-sheet path handling.
- Adds a publish-time confirmation dialog when a page is already scheduled if you try to either publish or schedule publish while the resource is still in queue to be schedule published.


auniverseaway (commented):
Small question on loading schedule stuff in da-title.

PR #838, opened 2026-03-12
feat: create Previewed version on preview

kptdobe:
- pages (html): create a "Previewed" version on each preview.
- sheets (json): create a "Previewed" version on each preview AND a "Published" version on each publish (was missing too).

Test: https://vpreview--da-live--adobe.aem.live/


chrischrischris (dismissed):

kptdobe:
sheets now auto create "Previewed" and "Published" (was not the case yet) versions.

PR #872, opened 2026-04-08
fix: fix block description icon

auniverseaway:
* Add try catch for fetching block details
* Fix icon id so it can be referenced from dom


auniverseaway:
image

asthabh23 (approved):

PR #877, opened 2026-04-17
feat: detect stale sheet/config and prevent concurrent overwrites

### LINKED ISSUE #704: [config] da-title web component should show save

auniverseaway:
* Add StaleCheck class to check for stale configs and sheets
  * Checks every 30 seconds for changes.
  * If changes, but nothing new locally, automatically updates local sheet.
  * If local and remote changes, present a dialog allowing to refresh or keep local edits. Keeping local edits will prevent auto saves and manual blue button saves.
* da-title updates to support the following:
  * disabledText property which doubles for disabling actions
  * Remove paper plane if only one action available
  * Any sheet inside `/.da/` can no longer be previewed or published.
  * Lazily check for disablePublish to ensure actions show ASAP.
  * Move `isSending` to class property to allow showing send event for save only actions.
  * Created a class-level event for save action

Resolves: #704

### Sheet
https://stale--da-live--adobe.aem.live/sheet#/aem-sandbox/block-collection/.da/temp-translate

### Config
https://stale--da-live--adobe.aem.live/config#/aem-sandbox/block-collection/

## Types of changes

### DIFF
--- a/blocks/edit/da-title/da-title.js
+++ b/blocks/edit/da-title/da-title.js
@@ -92,9 +92,7 @@ export default class DaTitle extends LitElement {
 
   setup() {
     this.reset();
-    this._actions = {
-      available: this.getAvailableActions()
-    }
+    this._actions = { available: this.getAvailableActions() };
     // Lazily filter the actions down
     this.filterActions();
   }
@@ -147,7 +145,7 @@ export default class DaTitle extends LitElement {
     ]);
 
     const { org, site, path, fullpath } = this.details;
-    
+
     // Only a valid path gets AEM-bound features
     if (path) {
       this._aemHrefs = await getAemHrefs({ path: fullpath });
--- a/blocks/edit/da-title/da-title.js
+++ b/blocks/edit/da-title/da-title.js
@@ -367,7 +367,7 @@ export default class DaTitle extends LitElement {
         @click=${() => this.handleAction(action)}
         class="con-button blue da-title-action"
         aria-label="Send"
-        data-popup-content="${this.disabledText ? this.disabledText : nothing}"
+        data-popup-content=${this.disabledText ?? nothing}
         ?disabled=${this.disabledText}>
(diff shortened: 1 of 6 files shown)


blocks/edit/da-title/da-title.js
```diff
@@ -207,14 +226,23 @@ export default class DaTitle extends LitElement {
...
     this._actions.open = false;
-    this.requestUpdate();
 
     const { org, site, view, fullpath, path } = this.details;
 
     const aemPath = `/${org}/${site}${path}`;
 
+    // Bail before writing if the remote drifted under us — protects against
+    // last-write-wins. Drift triggers the stale-content dialog via onStale.
+    if (view === 'sheet' || view === 'config') {
```
usman-khalid, line 238: I think this should only be for config sheets, considering the work in #779?

blocks/edit/da-title/da-title.js
```diff
@@ -337,7 +366,9 @@ export default class DaTitle extends LitElement {
       <button
         @click=${() => this.handleAction(action)}
         class="con-button blue da-title-action"
-        aria-label="Send">
+        aria-label="Send"
+        data-popup-content="${this.disabledText ? this.disabledText : nothing}"
```
usman-khalid:
```suggestion
        data-popup-content=${this.disabledText ?? nothing}
```

blocks/edit/da-title/da-title.js
auniverseaway: We don't know when this will be merged. We can remove this when that lands. There's currently a fairly major blocker that's been sitting for a month or two now.

usman-khalid (approved):

Issue #861, opened 2026-03-26
da-live deletion guards


chrischrischris:
1. Show how many files are about to be deleted: We can crawl the selected file/folder list and get a count to display to the user.  
2. Prevent deletion of 10,000 files at once.
3. Do not allow user to delete non-draft content if they do not have write permissions to .trash - This might be problematic though, and require consumers with complicated access rights to have to also rights to the .trash dir

kptdobe:
> 2. Prevent deletion of 10,000 files at once.

I am not sure who really needs to delete so many files. We might reduce this number to 50 for example with a message to contact the admin team.
A key shortcut could enable a force delete independent of the number of files (with a clear warning message).

> 3. Do not allow user to delete non-draft content if they do not have write permissions to .trash 

Or users should always have permissions to move to .trash.

PR #888, opened 2026-04-22
feat: Delete enhancements

chrischrischris:
Implement total count and max amount of items to delete allowed.

I upped the max delete amount to 1000 for now.  Or should we make it configurable?

Implements part of #861 

Screenshot 2026-04-22 at 3 58 28 PM
Screenshot 2026-04-28 at 3 34 34 PM
Screenshot 2026-04-28 at 3 34 59 PM

### DIFF
--- a/blocks/browse/da-list/da-list.js
+++ b/blocks/browse/da-list/da-list.js
@@ -15,6 +15,8 @@ const ICONS = [
   '/blocks/edit/img/Smock_Refresh_18_N.svg',
 ];
 
+const MAX_DELETE_COUNT = 1000;
+
 export default class DaList extends LitElement {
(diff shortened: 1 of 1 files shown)


kptdobe:
Re-thinking about it.. there are always scenarii where you really want to delete a folder with a lot of mess in it.

What about something like: 
- if less than 10, delete
- between 10 and 1000, user has to write "DELETE" to confirm this is really what they want to do
- contact admin if more than 1000

WDYT ?

auniverseaway (approved):
I think 1000 is fine.

auniverseaway:
I like @kptdobe suggestions.

PR #894, opened 2026-04-23
feat: Do not auto-create docs on edit nav

### LINKED ISSUE #702: [edit] Warn on 404
As a power user, I would like DA to prompt me to make a page if I visit `/edit` and the document does not exist so that I do not get ghost documents when I'm manipulating hashes instead of browsing the content tree.

chrischrischris:
Instead of automatically creating a doc when going to a new edit path, we now will present the user with a dialog asking what they would like to do.  If they choose to not create a doc, they are navigated to the containing directory for that file.

Screenshot 2026-04-23 at 3 38 26 PM

If a folder exists with the same name, they are also given the option of opening that folder.

Screenshot 2026-04-23 at 3 37 42 PM

fix #702

### DIFF
--- a/blocks/browse/da-new/da-new.js
+++ b/blocks/browse/da-new/da-new.js
@@ -7,6 +7,7 @@ import getEditPath from '../shared.js';
 const { default: getStyle } = await import(`${getNx()}/utils/styles.js`);
 const STYLE = await getStyle(import.meta.url);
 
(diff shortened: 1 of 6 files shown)


kptdobe (approved):

auniverseaway (approved):
If we don't already, we should disable the new button while we wait for the doc to be made.

PR #916, opened 2026-05-04
fix: prevent malformed API paths (double slashes, undefined segments, full URLs)

kptdobe:
## Summary

This PR fixes three distinct patterns of malformed API paths that caused 400 bad requests against da-admin.

### Fix 1 — Double slashes in API paths (`/list/org/repo//path`, etc.)

**File**: `blocks/shared/pathDetails.js`

**Root cause**: `details.fullpath` was set from the raw hash string, which preserved any `//` in the user's URL. `sanitizePathParts` strips double slashes for internal parsing but `details.fullpath` used the raw string, flowing into every API call.

**Fix**: Snapshot `pathParts` after the `isFolder` detection (before `getFullDetails` mutates the array via `.pop()`), then reconstruct the canonical path from those sanitized parts instead of the raw hash.

---

### Fix 2 — `/config//undefined/` requests

**File**: `blocks/shared/utils.js` — `fetchDaConfigs`

**Root cause**: No guard on `org` being falsy. If `org` is an empty string and `site` is a truthy `"undefined"` string, `fetchConfig(\`/${org}/${site}\`)` produced `fetchConfig("//undefined")` → URL `/config//undefined/`.

**Fix**: Return `[Promise.resolve(null)]` early when `org` is falsy, before any cache or fetch logic runs.

---

### Fix 3 — `/source/https://da.live` requests

**File**: `blocks/shared/utils.js` — `saveToDa`

**Root cause**: `saveToDa` constructed `${DA_ORIGIN}/source${path}` but `path` could arrive as a full URL from a caller that skipped relative-path extraction, producing a doubly-concatenated URL.

**Fix**: Validate `path` at entry — return `undefined` early when `path` is falsy, does not start with `/`, or contains `://`.

---

## Test plan

- [x] Fix 1: `getPathDetails` with a hash containing `//` returns a `fullpath` with no double slashes
- [x] Fix 2: `fetchDaConfigs({ org: '', site: 'something' })` returns early with `[null]` and makes no fetch calls
- [x] Fix 3: `saveToDa({ path: 'https://da.live' })` returns `undefined` without calling `daFetch`
- [x] All 1297 existing tests still pass

### DIFF
--- a/blocks/shared/pathDetails.js
+++ b/blocks/shared/pathDetails.js
@@ -148,6 +148,7 @@ export default function getPathDetails(loc) {
   const ext = getExtension(editor, pathParts.slice(-1)[0], isFolder);
 
   const depth = pathParts.length;
+  const cleanParts = [...pathParts];
 
   if (depth === 1) details = getOrgDetails({ editor, pathParts, ext });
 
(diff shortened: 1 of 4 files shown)


kptdobe:
Reopening from correctly named branch `fixpaths`

PR #921, opened 2026-05-06
fix: Remove trailing slash in edit mode

chrischrischris:


auniverseaway (approved):
Definitely more scope, but it would maybe have been better to just move edit.js to the newer hash detection stuff in NX.

PR #924, opened 2026-05-07
Experience Workspace

hannessolo:
https://ew--da-live--adobe.aem.live/drafts/hhertach/canvas?nx=ew-shell#/hannessolo/author-kit/demo

https://github.com/adobe/da-nx/pull/428


mhaack:
### Code review

Found 2 issues:

1. **No WebSocket auth-failure handler for 4401/4403 in canvas `prose.js`** — the new `WebsocketProvider` is created with no `connection-close` listener. PRs #937 and #943 landed exactly this fix in `blocks/edit/prose/index.js`: stop auto-reconnect (`shouldConnect = false`) and refresh the IMS token on a 4401 close code, bail entirely on 4403. Without it, an expired token during a canvas editing session will silently enter an infinite reconnect loop with no user feedback or recovery.

https://github.com/adobe/da-live/blob/90402e2c28492cd2ddaaa74cc4eb8eb36388ef0d/blocks/canvas/ew-editor-doc/prose.js#L84-L96

2. **IMS access token sent to extension iframes with `targetOrigin = '*'`** — `iframe-protocol.js` posts `{ ready: true, project, context: project, token }` using `'*'` as the target origin. Any page that ends up loaded in that iframe position (including one navigated to a different origin) will receive the bearer token. The WYSIWYG iframe path in the same PR correctly derives the origin via `new URL(iframe.src).origin` and uses that instead.

https://github.com/adobe/da-live/blob/90402e2c28492cd2ddaaa74cc4eb8eb36388ef0d/blocks/canvas/ew-panel-extensions/iframe-protocol.js#L77-L84

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

blocks/canvas/editor-utils/command-defs.js
mhaack, line 1: Isn't this something we can reuse from the todays editor?

mhaack (dismissed):
Next to the Claude review comment this LGTM, lets get the V1 in.

hannessolo:
Re: claude review:

Issue 1 is fixed.

Issue 2 we won't fix here, as it's also an issue with the "old" editor's extension loading implementation.

blocks/canvas/editor-utils/command-defs.js
hannessolo: Today's editor uses a different format, which is integrated with prosemirror. IMO it would be more effort than it's worth.

auniverseaway (dismissed):

mhaack (dismissed):

mhaack (approved):

PR #933, opened 2026-05-13
fix: coerce non-string da-metadata values to avoid "[object Object]"

chrischrischris:
## Summary

- `setDaMetadata` is the single boundary into the `daMetadata` Y.Map, but it forwarded values verbatim. `doc2aem` later interpolates each value into `<div>${value}</div>`, so any non-string entry (object, Y type) ends up serialized as the literal string `[object Object]` in saved HTML — observed in the wild as:
  ```html
  <div class="da-metadata"><div><div>acceptedhashes</div><div>[object Object]</div></div><div><div>rejectedhashes</div><div>[object Object]</div></div></div>
  ```
- Coerce at the boundary: stringify objects via `JSON.stringify`, other non-strings via `String()`, and `console.warn` so the offending caller is visible in dev tools next time it happens.
- Added two unit tests covering object and primitive coercion.

## Notes for reviewers

- I couldn't pinpoint which call site is producing non-string values. Every in-tree caller (`addToHashMetadata` and the version-restore loop in `da-editor.js`) already passes strings, and the lowercase keys (`acceptedhashes`/`rejectedhashes`) in the observed HTML don't match our `ACCEPTED_KEY`/`REJECTED_KEY` constants (`acceptedHashes`/`rejectedHashes`) — so the corrupt entries likely came from outside `setDaMetadata` entirely. The new `console.warn` should help surface the source.
- A companion defensive fix is also worth doing in `da-parser`'s `doc2aem` ([parser.js:753](https://github.com/adobe/da-tools/blob/main/da-parser/src/doc/parser.js#L753)), which neither stringifies nor HTML-escapes the keys/values. Happy to put up a PR there as a follow-up.

## Test plan

- [x] `npx web-test-runner test/unit/blocks/edit/utils/helpers.test.js` — 64 passing, including new coercion tests.
- [ ] Reproduce the original symptom (a corrupted doc with `[object Object]` in `da-metadata`), open it in da-live, edit, save, and confirm no new `[object Object]` rows are written.


chrischrischris:
Superseded by #1046. The code is unchanged — this PR's branch name (`claude/musing-shaw-35684b`) violates the IMS branch spec (`/^[a-zA-Z0-9-]{1,8}$/`), which made `ims.setup` fail and blocked the entire e2e suite plus preview-env auth. Reopened on the compliant branch `metafix`.

PR #935, opened 2026-05-13
add fallback for path sanitization logic

usman-khalid:


chrischrischris:
Replacing with https://github.com/adobe/da-live/pull/936

PR #941, opened 2026-05-15
feat: Auth state management: in-page modal + cross-tab propagation

chrischrischris:
<html><head></head><body>
<h2>Why</h2>
<p>Three related gaps in how da-live handles auth state changes:</p>
<ol>
<li><strong>Token failures used to navigate the user away.</strong> A <code>daFetch</code> 401 or an unrecoverable collab <code>4401</code> close called <code>handleSignIn()</code>, which redirects to IMS. Any unsaved editor state was lost in the round-trip.</li>
<li><strong>Cross-tab sign-out wasn't propagating.</strong> With nx2's <code>handleSignOut</code> (which doesn't touch <code>nx-ims</code>), signing out in Tab A left Tab B's editor fully functional. The user could keep typing — and da-collab kept persisting Tab B's edits to da-admin using the bearer it cached at WS handshake, since access tokens are stateless and survive IMS sign-out until their natural expiry.</li>
<li><strong>No coordinated UX for "session lost".</strong> Different code paths handled auth loss differently (or not at all): daFetch redirected to IMS, prose silently retried, etc.</li>
</ol>
<h2>What</h2>
<ul>
<li><strong><code>&lt;da-auth-banner&gt;</code> modal.</strong> Native <code>&lt;dialog&gt;</code> with <code>showModal()</code> that blocks page interaction; "Sign in" button triggers the standard IMS flow. Mounted via <code>showAuthBanner()</code> (idempotent — second call returns the existing instance). Replaces what used to be a redirect.</li>
<li><strong>Cross-tab auth monitor.</strong> Single <code>storage</code> event listener (attached as a side-effect of the first <code>initIms()</code> call) that re-reads <code>window.adobeIMS.getAccessToken()</code> on every change. Works under both nx and nx2 because it inspects live imslib state instead of keying off <code>nx-ims</code>. On auth loss → mount the modal + <code>wsProvider.disconnect()</code> (so da-collab can't keep authoring writes with Tab B's cached bearer). On auth gain → <code>window.location.reload()</code> to pick up the new session cleanly.</li>
<li><strong><code>daFetch</code> refresh-and-retry on 401.</strong> Before any user-visible disruption, ask imslib to refresh the token and retry the failed request once. Rescues the cross-tab "just signed in, but this tab hasn't caught up yet" race. Falls through to the modal only if there's truly no token.</li>
<li><strong><code>prose/index.js</code> 4401 handling.</strong> When the collab WS reconnect handshake gets <code>4401 'auth'</code> from da-collab (new in da-collab#149), refresh the token and retry. If unrecoverable and the user had a token, mount the modal.</li>
<li><strong>Sign-out lands on home.</strong> <code>scripts.js</code> inspects the IMS callback URL (<code>old_hash=</code> present, <code>access_token=</code> absent → sign-out return) and <code>window.location.replace('/')</code> before <code>loadArea()</code> runs, so the sign-out tab doesn't briefly mount the editor + modal for the doc the user just left.</li>
</ul>
<h2>Trust model notes</h2>
<p>da-collab's auth check is <strong>handshake-only</strong> — <code>webSocket.auth</code> is cached and never re-verified. The only durable protections against post-sign-out writes are:</p>
<ul>
<li><strong>Server-side</strong> (<a href="https://claude.ai/da-collab/src/shareddoc.js#L368">da-collab/shareddoc.js:368</a>): when da-admin returns 401 on a persist, da-collab calls <code>closeConn</code> on every connection in the room. So token rotation past expiry self-heals — every tab gets force-reconnected with whatever token <code>getAuthToken()</code> reads live (which imslib has rotated by then). This is what makes 24+ hour sessions work transparently.</li>
<li><strong>Client-side</strong> (this PR): the auth monitor calls <code>wsProvider.disconnect()</code> the moment cross-tab sign-out is detected, so Tab B drops out of the room before any further yjs updates can flow through its still-authed connection.</li>
</ul>


auniverseaway (commented):
1. Looks like there's some test failurs.
2. I'd make sure to use da-dialog if possible.

auniverseaway (approved):

PR #955, opened 2026-05-21
fix: bound y-websocket reconnect rate on transient post-open closes

kptdobe:
## Summary

- y-websocket resets its `wsUnsuccessfulReconnects` counter on every successful `onopen`, so any close that follows a brief successful handshake reschedules at ~100 ms. The 4401/4403 guard from #943 does not cover non-auth paths (1011 from `initSession` catch, 1005 from `closeConn`, 1006 from socket reset), and single users behind corporate Zscaler proxies have sustained 5k+ WS upgrades/sec/IP because of it (see COR-43).
- Add a rapid-reconnect guard to `createConnection`: track open-then-close intervals on the provider and apply manual exponential backoff (1s/2s/4s/… capped at `SHORT_SESSION_MAX_MS = 30 s`) when sessions shorter than `MIN_HEALTHY_SESSION_MS = 5 s` repeat. A healthy session (≥ 5 s lived) resets the counter so a routine reconnect after a long session still reconnects via y-websocket's own timer.
- The auth path from #943 is unchanged and the new guard never increments on 4401/4403 closes — the auth flow has its own loop guard via `lastSentToken`.

## Implementation notes

- The new constants live as module-private consts in `blocks/edit/prose/index.js`.
- The guard sits BEFORE the existing token-refresh `provider.protocols` reassignment so a token refresh still happens for the next manual `provider.connect()` after the backoff.
- y-websocket's internal 100 ms `setTimeout(setupWS)` still fires from `onclose`, but `provider.disconnect()` flips `shouldConnect = false`, so that timer's `setupWS` call no-ops. The manual `setTimeout` is what re-arms the connection.
- No changes to da-y-wrapper or upstream y-websocket — the fix sits at the consuming-call site so it ships with da-live's normal release cadence.

## Test plan

- [x] `npm run lint` clean on the touched files
- [x] Focused unit tests pass: `npx wtr "./test/unit/blocks/edit/prose/index.test.js"` — 35/35 green, including the new `prose/index createConnection rapid-reconnect guard (COR-44)` suite (7 new scenarios) plus an updated existing test that now reflects the post-guard `shouldConnect = false` parking state.
- [ ] After merge, query Coralogix and confirm the dominant da-collab URLs from the COR-1 daily review drop out of the >95% exception-ratio bucket and that the populated `"Network connection lost."` exception count returns to baseline (<10k/24h) within 48 h.

Test: https://cor44--da-live--adobe.aem.live/

## Refs

- Issue: COR-44 (rapid-reconnect guard)
- Parent: COR-43 (investigation that surfaced the y-websocket reset behavior)
- Prior PR: #943 (auth-close 4401/4403 guard)
- Supersedes: #954 (same change on a non-compliant branch name)

### DIFF
--- a/blocks/edit/prose/index.js
+++ b/blocks/edit/prose/index.js
@@ -29,6 +29,16 @@ import { getDiffClass, checkForLocNodes, addActiveView } from './diff/diff-utils
 import { debounce, initDaMetadata } from '../utils/helpers.js';
 import { forceSave } from './forcesave.js';
 
+// Rapid-reconnect guard (COR-44): y-websocket resets its backoff counter on
+// every successful onopen, so a close that follows a brief successful
+// handshake reschedules at 100ms. Single users behind corporate proxies can
+// sustain thousands of WS upgrades/sec/IP through this path. The guard below
+// detects open-then-close cycles shorter than MIN_HEALTHY_SESSION_MS and
+// applies a manual exponential backoff before letting the provider reconnect.
+const MIN_HEALTHY_SESSION_MS = 5000;
+const SHORT_SESSION_BASE_MS = 1000;
+const SHORT_SESSION_MAX_MS = 30000;
+
 async function checkDoc(path) {
   return daFetch(path, { method: 'HEAD' });
 }
@@ -56,6 +66,13 @@ export async function createConnection(path) {
   provider.maxBackoffTime = 30000;
 
   let lastSentToken = token || null;
+  let lastOpenAt = 0;
+  let failedShortSessions = 0;
+
+  provider.on('status', (st) => {
+    if (st?.status === 'connected') lastOpenAt = Date.now();
+  });
+
   provider.on('connection-close', async (event) => {
     if (event?.code === 4401 || event?.code === 4403) {
       provider.shouldConnect = false;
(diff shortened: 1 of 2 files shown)


blocks/edit/prose/index.js
```diff
@@ -78,6 +95,28 @@ export async function createConnection(path) {
       provider.connect();
       return;
     }
+
+    // Non-auth close: rapid-reconnect guard. y-websocket's own 100ms
+    // setTimeout(setupWS) still fires from onclose, but provider.disconnect()
+    // flips shouldConnect=false so that timer's setupWS call no-ops. The
+    // manual setTimeout below is what re-arms the connection.
+    const sessionMs = lastOpenAt ? Date.now() - lastOpenAt : 0;
+    if (sessionMs >= MIN_HEALTHY_SESSION_MS) {
```
chrischrischris, line 105: Should `lastOpenAt` be reset to `0` here?

chrischrischris (dismissed):

blocks/edit/prose/index.js
kptdobe: Nice catch. Fixed and added the corresponding test!

PR #956, opened 2026-05-21
feat: preview/publish single + bulk on browse

sharanyavinod:
## Description

- Introduces Preview/Publish in the action bar on single and multiple files selection
- Follows same pattern as delete for the confirm dialog for initial confirm -> switches then to load within action bar + clickable toast to show the urls if need, which auto closes. UX is rudimentary for now, to be revisited when implementing the new designs.
- For publish, if a subset of the selections have a publish scheduled, user is asked to confirm overriding - all or nothing atm, can iterate if needed to give more granular control
- Extracts reusable logic from the actions within the edit view

Preview URL: https://ew-pub--da-live--adobe.aem.live/#/exp-workspace/frescopa

## Motivation and Context

- Users can preview directly from browse without stepping into edit and can preview multiple files easily

## How Has This Been Tested?

- Introduced unit tests consistent with existing tests

Screenshot 2026-05-21 at 16 08 32

## Types of changes

## Checklist:


mhaack:
### Code review

Found 3 issues:

1. **Wrong permission role requested when preview fails during a publish action** — `aemAction` in `shared/utils.js` calls `saveToAem(path, 'preview')` first, then returns the preview error to `da-title.js`. But `da-title.js` calls `this.handleError(json, action)` where `action` is always `'publish'`, overwriting `json.error.action` via object spread: `this._status = { ...json.error, action }`. When a user lacks preview access and tries to publish, clicking "Request access" will request the `basic_publish` role instead of the `basic_author` role needed to preview.

https://github.com/adobe/da-live/blob/c25faf2958f3a9cf1afed60c651df5ae4592fa9f/blocks/edit/da-title/da-title.js#L279-L283

2. **Toast button and error panel render simultaneously on partial success** — In `runAemQueue`, when at least one item succeeds, `_aemActionState = { verb, results }` is set and then `handleClear()` is called. `handleClear()` does not invoke `handleErrorClose()`, so `_itemErrors` retains any failed items. The render template shows the error panel whenever `!this._confirm && this._itemErrors.length`, and the toast button whenever `this._aemActionState?.results`. After a partial success, both conditions are true at the same time, causing both UI elements to appear simultaneously.

https://github.com/adobe/da-live/blob/c25faf2958f3a9cf1afed60c651df5ae4592fa9f/blocks/browse/da-list/da-list.js#L537-L546

3. **`getExistingSchedule` is duplicated instead of reused** — The PR adds `getExistingSchedule` (and `SNAPSHOT_SCHEDULER_URL`) to `blocks/shared/utils.js`, but an identical implementation already exists in `blocks/edit/da-prepare/actions/scheduler/utils.js`. The PR should import from the existing location rather than copying it.

New copy: https://github.com/adobe/da-live/blob/c25faf2958f3a9cf1afed60c651df5ae4592fa9f/blocks/shared/utils.js#L188-L200

Existing copy: https://github.com/adobe/da-live/blob/c25faf2958f3a9cf1afed60c651df5ae4592fa9f/blocks/edit/da-prepare/actions/scheduler/utils.js#L24-L36

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

mhaack:
Some more testing feedback:

* preview or publish > once done click on the blue confirmation toast bubble to see the dialog
> after few seconds, when the blue confirmation is removed the dialog is closed as well, but it should stay

* when selection a folder & some pages (see screenshot) the dialog should get an extra message about the folder not being published
Screenshot 2026-06-01 at 08 23 41

sharanyavinod:
@mhaack Thanks, fixed both of them.

chrischrischris (commented):
Nice!  I do wonder if we can move some of the code out of da-list into a file we dynamically import when preview/publish is started.

sharanyavinod:
> Nice! I do wonder if we can move some of the code out of da-list into a file we dynamically import when preview/publish is started.

Indeed - was thinking of handling this in a follow-up PR since we also need to switch the styling to nx2.

mhaack (approved):

PR #960, opened 2026-05-22
fix: update toolbar position + open on focus

sharanyavinod:
## Description

- Display toolbar on focus
- Position toolbar always at bottom of page instead of anchored to the selection

Preview: https://ew-tlbr--da-live--adobe.aem.live/canvas#/exp-workspace/frescopa/index

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/blocks/canvas/ew-editor-wysiwyg/utils/handlers.js
+++ b/blocks/canvas/ew-editor-wysiwyg/utils/handlers.js
@@ -1,8 +1,6 @@
 import { TextSelection, yUndo, yRedo } from 'da-y-wrapper';
 import {
   getSelectionToolbar,
-  hideSelectionToolbar,
-  TOOLBAR_PADDING_GAP,
   NX_QUICK_EDIT_IFRAME_SELECTION_META,
   NX_QUICK_EDIT_CLEAR_IFRAME_SELECTION_ORIGIN_META,
 } from '../../editor-utils/selection-toolbar.js';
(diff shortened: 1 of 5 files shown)


mhaack (dismissed):

blocks/canvas/ew-editor-wysiwyg/utils/handlers.js
```diff
@@ -157,5 +154,5 @@ export function handleIframeSelectionChange(data, ctx) {
     return;
   }
   if (!handleSelectionChange(data, ctx, { fromQuickEditIframe: true })) return;
-  positionSelectionToolbarFromIframe(data, ctx);
```
hannessolo, line 160: Maybe rename this to just "showToolbarIniFrame" or similar

hannessolo (commented):
I think there might be a scroll event listener somewhere that repositions the toolbar in the wysiwyg frame on scroll - unless I already removed that?

I can check later or maybe you can find it. But we probably don't need it anymore if we have the position be fixed.

hannessolo:
I like this. It also paves the way for the table controls (add/remove row) that I was sort of struggling to place somewhere in the UI. Now it can live in the toolbar.

hannessolo (changes requested):
I found a bug. The state of the toolbar becomes out of sync - eg this text selected is clearly not italic.

Image

sharanyavinod:
> I think there might be a scroll event listener somewhere that repositions the toolbar in the wysiwyg frame on scroll - unless I already removed that?
> 
> I can check later or maybe you can find it. But we probably don't need it anymore if we have the position be fixed.

Thanks for the callout. Cleaned it up.

sharanyavinod:
> I found a bug. The state of the toolbar becomes out of sync - eg this text selected is clearly not italic.

Nice catch, fixed it.

hannessolo (approved):

PR #963, opened 2026-05-26
fix: metadata rendering in layout mode

mhaack:
We render the metadata block visible in the layout mode. This is not how it should be.

Test:

* https://ew-meta--da-live--adobe.aem.live/canvas#/exp-workspace/frescopa/index

### DIFF
--- a/blocks/canvas/editor-utils/editor-utils.js
+++ b/blocks/canvas/editor-utils/editor-utils.js
@@ -169,24 +169,21 @@ export function getInstrumentedHTML(view) {
   const originalTables = view.dom.querySelectorAll('table');
   const clonedTables = editorClone.querySelectorAll('table');
   clonedTables.forEach((table, index) => {
-    // for metadata table skip marker instrumentation and hide the table
     const firstRow = table.querySelector('tr');
     const firstCellText = firstRow?.cells?.[0]?.textContent?.trim().toLowerCase();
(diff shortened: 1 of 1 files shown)


blocks/canvas/editor-utils/editor-utils.js
```diff
@@ -169,17 +169,21 @@ export function getInstrumentedHTML(view) {
...
-    blockMarker.className = 'block-marker';
-    try {
-      const position = view.posAtDOM(originalTables[index], 0);
-      blockMarker.setAttribute('data-prose-index', position);
-    } catch (e) {
-      // eslint-disable-next-line no-console
-      console.warn('Could not find position for table block:', e);
+    const firstRow = table.querySelector('tr');
+    const firstCellText = firstRow?.cells?.[0]?.textContent?.trim().toLowerCase();
+    if (firstCellText !== 'metadata') {
```
hannessolo: IMO `if (firstCellText === 'metadata') return;` would be cleaner but this works too.

hannessolo (dismissed):

blocks/canvas/editor-utils/editor-utils.js
mhaack: Good point, agreed. Updated.

hannessolo (approved):

sharanyavinod (approved):

PR #966, opened 2026-05-26
fix: reload loop with nested lists in wysiwyg

hannessolo:
https://ewfl--da-live--adobe.aem.live/canvas#/scdemos/demo/drafts/chelms/index

Fixes page with nested lists


helms-charity (approved):
My test page is looking good now, thanks!

sharanyavinod (approved):

PR #976, opened 2026-05-29
fix: block library missing title for grouped blocks

mhaack:
Test:
* https://ew-lib--da-live--adobe.aem.live/canvas#/exp-workspace/frescopa/drafts/mhaack/lib-test

fixes: https://jira.corp.adobe.com/browse/SITES-45828

Before:
Screenshot 2026-05-29 at 11 51 35

After:
Screenshot 2026-05-29 at 11 50 36

### DIFF
--- a/blocks/canvas/ew-panel-extensions/helpers.js
+++ b/blocks/canvas/ew-panel-extensions/helpers.js
@@ -135,9 +135,11 @@ function getLibraryMetadata(el) {
 
 function transformBlock(block) {
   const prevSib = block.previousElementSibling;
-  const item = isHeading(prevSib) && prevSib.textContent
-    ? { name: prevSib.textContent }
-    : getBlockName(block.className || '');
+  const item = block.dataset.groupheading
+    ? { name: block.dataset.groupheading }
+    : isHeading(prevSib) && prevSib.textContent
+      ? { name: prevSib.textContent }
+      : getBlockName(block.className || '');
   item.dom = block.dataset?.isgroup ? processGroupBlock(block) : getBlockTableHtml(block);
 
   const metaEl = block.nextElementSibling?.classList.contains('library-metadata')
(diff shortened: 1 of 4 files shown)


blocks/canvas/ew-panel-extensions/helpers.js
```diff
@@ -135,9 +135,14 @@ function getLibraryMetadata(el) {
 
 function transformBlock(block) {
   const prevSib = block.previousElementSibling;
-  const item = isHeading(prevSib) && prevSib.textContent
-    ? { name: prevSib.textContent }
-    : getBlockName(block.className || '');
+  let item;
+  if (block.dataset.groupheading) {
```
anfibiacreativa, line 139: maybe worth adding a test for the three branches in transformBlock (groupheading attr, heading sibling, fallback to getBlockName). without it the same thing can regress silently.

blocks/canvas/ew-panel-extensions/helpers.js
```diff
@@ -135,9 +135,14 @@ function getLibraryMetadata(el) {
...
+    item = { name: block.dataset.groupheading };
```
anfibiacreativa, line 140: is this data-groupheading (one word) or data-group-heading? the dataset key needs to match exactly.

mhaack: This is correct:

```
block
<div data-isgroup=​"true" data-groupheading=​"Hero with text">​…​</div>​

block.dataset
DOMStringMap {isgroup: 'true', groupheading: 'Hero with text'}
```

blocks/canvas/ew-panel-extensions/helpers.js
mhaack: Yes I agree, added tests

chrischrischris (approved):

PR #990, opened 2026-06-08
feat(ew): add mod-k keyboard shortcut for link editing, refactor link dialog

mhaack:
## Description

Add Mod-k link editing keyboard shortcut to EW canvas

Pressing Mod-k in the EW Canvas editor now opens the link dialog, matching the DA editor's existing behaviour.

  - Extracts the inline link dialog from EwSelectionToolbar into a new reusable blocks/shared/da-link-dialog Lit component
  - Fixes findLinkInRange to detect a link when the cursor is inside it (empty selection), enabling pre-fill of URL and display text
  - Wires Mod-k keymap entry in prose.js; guarded against read-only editors
  - Adds unit tests for da-link-dialog (10) and findLinkInRange (6)

Test:
  - Url: https://akey--da-live--adobe.aem.live/canvas#/exp-workspace/frescopa/drafts/mhaack/something-not-here

## Related Issue

https://jira.corp.adobe.com/browse/SITES-46088


blocks/shared/da-link-dialog/da-link-dialog.js
```diff
@@ -0,0 +1,87 @@
...
+    if (e.key === 'Escape') {
+      e.stopPropagation();
+      this._onCancel();
+    }
+  }
+
+  render() {
+    if (!this.open) return nothing;
+    return html`
+      <div class="link-dialog"
```
sharanyavinod: Why dont we use native dialog here so we get accessibility and other native dialog behaviors ootb?

On that note, why don't we use the shared dialog in da-nx? The required body can be passed in, and all other dialog specific behaviours would be handled there.

sharanyavinod: I see the dialog rendering was pre-existing - but maybe this would be a good opportunity to swap out to use the shared dialog?

blocks/shared/da-link-dialog/da-link-dialog.js
```diff
@@ -0,0 +1,87 @@
...
+        @click=${this._onBackdropClick}
+        @keydown=${this._onBackdropKeydown}>
+        <form class="link-form" @submit=${this._onSubmit}>
+          <label class="link-form-field">
+            <span>URL</span>
+            <input name="link-href" type="text" placeholder="https://…"
```
sharanyavinod: type url previously is switched to text here -> intentional?

mhaack: good catch, fixed

blocks/shared/da-link-dialog/da-link-dialog.js
mhaack: Good idea, change the dialog to `nx-dialog` but kept `da-link-dialog.js` as a wrapper to have shortcuts in prose and toolbar decoupled.

sharanyavinod (dismissed):

mhaack:
FYI: change the URL input type back to `text` and added `if (/^(javascript|data|vbscript):/i.test(href)) return;` for validation as input type `url` only allows for absolut URLs but we need to support relative paths as well.

blocks/shared/da-link-dialog/da-link-dialog.js
```diff
@@ -0,0 +1,73 @@
...
+    super.connectedCallback();
+    this.shadowRoot.adoptedStyleSheets = [styles];
+  }
+
+  _onSubmit(e) {
+    e.preventDefault();
+    const form = e.target;
+    const href = form.elements['link-href'].value.trim();
+    if (!href) return;
+    if (/^(javascript|data|vbscript):/i.test(href)) return;
```
sharanyavinod: Instead of failing silently, maybe we could use validation apis of input fields to give feedback to user?

mhaack: added some validation

hannessolo (approved):

PR #995, opened 2026-06-10
feat(ew): read default canvas editor view from site config

mhaack:
## Summary

- Adds support for an `ew.canvasDefaultView` flag in the site DA config to control which view (`content`, `split`, or `layout`) the canvas opens with on first visit
- Session storage (persisted user preference) always takes precedence — the config is only consulted when no persisted value exists
- Invalid or missing flag values fall back to `layout` (existing default)
- Added tests

Test
- https://vwcfg--da-live--adobe.aem.live/canvas#/exp-workspace/frescopa/coffee

## How to configure

Add a flag entry to the site config:

| key | value |
|-----|-------|
| `ew.canvasDefaultView` | `content` \| `split` \| `layout` |

## Test plan

- [ ] Open canvas with no session storage and no config flag → lands on `layout`
- [ ] Set `ew.canvasDefaultView = content` in site config, clear session storage → lands on `content`
- [ ] Set `ew.canvasDefaultView = split` in site config, clear session storage → lands on `split`
- [ ] Set an invalid flag value → falls back to `layout`
- [ ] Switch view manually → persists to session storage; config flag no longer applies on next load

### DIFF
--- a/blocks/canvas/utils/view.js
+++ b/blocks/canvas/utils/view.js
@@ -15,10 +15,12 @@ export async function readInitialCanvasEditorView({ org, site }) {
   } catch { /* ignore if browser disallows session storage */ }
 
   try {
-    const siteConfig = await fetchDaConfigs({ org, site })[1];
+    const [, siteConfig] = await Promise.all(fetchDaConfigs({ org, site }));
     const flag = siteConfig?.flags?.data?.find((f) => f.key === 'ew.canvasDefaultView');
     if (flag) return normalizeCanvasEditorView(flag.value);
-  } catch { /* ignore config fetch errors */ }
+  } catch (e) {
+    if (!(e instanceof TypeError) && !(e instanceof SyntaxError)) throw e;
+  }
 
   return 'layout';
 }
--- a/blocks/canvas/utils/view.js
+++ b/blocks/canvas/utils/view.js
@@ -1,6 +1,6 @@
 import { fetchDaConfigs } from '../../shared/utils.js';
 
-export const CANVAS_EDITOR_VIEW_KEY = 'nx-canvas-editor-view';
+const CANVAS_EDITOR_VIEW_KEY = 'nx-canvas-editor-view';
 
 export function normalizeCanvasEditorView(view) {
   if (view === 'content') return 'content';
--- a/blocks/canvas/utils/view.js
+++ b/blocks/canvas/utils/view.js
@@ -1,0 +1,30 @@
+import { fetchDaConfigs } from '../../shared/utils.js';
+
+export const CANVAS_EDITOR_VIEW_KEY = 'nx-canvas-editor-view';
+
+export function normalizeCanvasEditorView(view) {
+  if (view === 'content') return 'content';
+  if (view === 'split') return 'split';
+  return 'layout';
+}
+
+export async function readInitialCanvasEditorView({ org, site }) {
+  try {
+    const persisted = sessionStorage.getItem(CANVAS_EDITOR_VIEW_KEY);
+    if (persisted) return normalizeCanvasEditorView(persisted);
+  } catch { /* ignore if browser disallows session storage */ }
+
+  try {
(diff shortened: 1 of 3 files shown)


blocks/canvas/utils/view.js
```diff
@@ -0,0 +1,30 @@
...
+
+export async function readInitialCanvasEditorView({ org, site }) {
+  try {
+    const persisted = sessionStorage.getItem(CANVAS_EDITOR_VIEW_KEY);
+    if (persisted) return normalizeCanvasEditorView(persisted);
+  } catch { /* ignore if browser disallows session storage */ }
+
+  try {
+    const siteConfig = await fetchDaConfigs({ org, site })[1];
+    const flag = siteConfig?.flags?.data?.find((f) => f.key === 'ew.canvasDefaultView');
```
anfibiacreativa, line 19: why not destructure? const [, siteConfig] = await Promise.all(fetchDaConfigs({ org, site })) at least makes the intent readable. the [1] magic number will confuse the next person and fail invisibly if the shape shifts.

test/unit/blocks/canvas/utils/view.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+  // fetchDaConfigs caches by org/site — use a unique pair per test to avoid cache hits.
+  function ctx() { return { org: `org-${testIndex}`, site: `site-${testIndex}` }; }
+
+  function mockConfig(flags) {
+    window.fetch = () => Promise.resolve(new Response(JSON.stringify({ flags }), { status: 200 }));
+  }
+
+  function mockConfigError() {
+    window.fetch = () => Promise.resolve(new Response('', { status: 500 }));
```
anfibiacreativa: can we verify what fetchDaConfigs does with a 500? if it doesn't throw, this test is testing nothing about the catch branch. worth checking or adjusting the mock to actually throw.

blocks/canvas/utils/view.js
```diff
@@ -0,0 +1,30 @@
...
+    if (flag) return normalizeCanvasEditorView(flag.value);
+  } catch { /* ignore config fetch errors */ }
+
+  return 'layout';
+}
+
+export function persistCanvasEditorView(view) {
+  try {
```
anfibiacreativa, line 29: these are exported functions with real branches. we should have a describe('normalizeCanvasEditorView') block and at least a quick test for persistCanvasEditorView (writes to sessionStorage, normalizes the value before writing).

blocks/canvas/utils/view.js
```diff
@@ -0,0 +1,30 @@
...
+    const persisted = sessionStorage.getItem(CANVAS_EDITOR_VIEW_KEY);
+    if (persisted) return normalizeCanvasEditorView(persisted);
+  } catch { /* ignore if browser disallows session storage */ }
+
+  try {
+    const siteConfig = await fetchDaConfigs({ org, site })[1];
+    const flag = siteConfig?.flags?.data?.find((f) => f.key === 'ew.canvasDefaultView');
+    if (flag) return normalizeCanvasEditorView(flag.value);
+  } catch { /* ignore config fetch errors */ }
```
anfibiacreativa: maybe only catch NetworkError or check response status and let real programming errors bubble. silencing all errors here makes this a pain to debug.

anfibiacreativa (changes requested):

test/unit/blocks/canvas/utils/view.test.js
mhaack: Yes good catch I will remove it. I also don' really want to test the fetchDaConfigs here this should be seperated

blocks/canvas/utils/view.js
mhaack: done

blocks/canvas/utils/view.js
mhaack: done, test added

anfibiacreativa (approved):

sharanyavinod (approved):

PR #1008, opened 2026-06-17
fix: strip .html extension when unpublishing from browse

### LINKED ISSUE #1007: Unpublish sends path with .html extension
While this was accepted by the admin API (different issue) and the `.md` was removed from EDS, it did not purge the cache correctly, so users were still seeing their unpublished page instead of a 404.

rofe:
Fix #1007

Test URL: https://unpubext--da-live--adobe.aem.live/

### DIFF
--- a/test/fixtures/nx/public/utils/tree.js
+++ b/test/fixtures/nx/public/utils/tree.js
@@ -1,7 +1,13 @@
 export class Queue {
-  constructor() { this.items = []; }
+  constructor(callback) {
+    this.callback = callback;
+    this.items = [];
+  }
 
-  push(item) { this.items.push(item); }
+  push(item) {
+    this.items.push(item);
+    return this.callback ? this.callback(item) : undefined;
+  }
 
   shift() { return this.items.shift(); }
 }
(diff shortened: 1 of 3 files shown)


test/fixtures/nx/public/utils/tree.js
```diff
@@ -1,7 +1,13 @@
...
-  constructor() { this.items = []; }
+  constructor(callback) {
+    this.callback = callback;
+    this.items = [];
+  }
 
-  push(item) { this.items.push(item); }
+  push(item) {
+    this.items.push(item);
+    return this.callback ? this.callback(item) : undefined;
```
rofe, line 9: This fixture `Queue` was previously a no-op — `push()` enqueued items but never invoked a callback. The real nx `Queue` runs the callback per item (`new Queue(callback, …)`), which is how `handleConfirmDelete` drives its per-item unpublish logic.

Without this change the new test can't reach the `.html`-stripping code path being fixed here (the callback never fires). I kept the stub minimal — just callback invocation — and left out the real `Queue`'s concurrency/throttle/on-complete behavior since the test doesn't need it.

chrischrischris (approved):

PR #1015, opened 2026-06-18
fix: recover from image upload failures in the editor

kptdobe:
## Summary

- When `daFetch` throws or returns a non-ok response during image drag-and-drop upload, the FPO placeholder (`fpo.svg`) is now removed from the document instead of being left stuck permanently
- A red toast notification is shown to the author: _"Failed to upload "filename". Please try again."_ — auto-dismisses after 5 s
- If the upload succeeds but the content URL fails to render (broken CDN, slow propagation), the placeholder is still replaced with the real URL so it isn't stuck as `fpo.svg` either

## Root cause

On Windows, dragging an image from File Explorer into the editor can cause the upload `PUT` request to fail (browser navigation to a `file:///` URI competing with the upload, or corporate proxy blocking requests to the `.grs` path). The original code had no error recovery — any failure silently left the FPO placeholder in the document, and since it was already synced to the Yjs collab layer, it persisted across reloads.

## Test plan

Test page: https://imgdrop--da-live--adobe.aem.page

- [ ] Drag and drop an image on Mac/Chrome — image uploads and replaces the FPO as before
- [ ] Simulate upload failure (DevTools → block the PUT request) — FPO is removed, toast appears
- [ ] All unit tests pass (`npm test`)

### DIFF
--- a/blocks/browse/da-browse/da-browse.js
+++ b/blocks/browse/da-browse/da-browse.js
@@ -141,7 +141,8 @@ export default class DaBrowse extends LitElement {
       <da-new
         @newitem=${this.handleNewItem}
         fullpath="${this.details.fullpath}"
-        editor="${this.editor}">
+        editor="${this.editor}"
+        ?folderOnly=${this.isRootFolder(this.details.fullpath)}>
       </da-new>`;
   }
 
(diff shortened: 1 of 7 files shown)


kptdobe:
Validating with customer.

PR #1018, opened 2026-06-19
refactor toolbar in ew

hannessolo:
https://ewtb--da-live--adobe.aem.live

### DIFF
--- a/blocks/canvas/editor-utils/selection-toolbar-controller.js
+++ b/blocks/canvas/editor-utils/selection-toolbar-controller.js
@@ -1,0 +1,93 @@
+/**
+ * Owns the selection-toolbar's visibility state across both editor modes (doc
+ * and wysiwyg). Each editor declares activation/deactivation explicitly rather
+ * than the toolbar introspecting `view.hasFocus()` — which lies in wysiwyg mode
+ * (see editor-doc cursor-broadcast hack in handlers.js).
+ */
+
+const listeners = new Set();
+
+const state = {
+  mode: 'none', // 'doc' | 'wysiwyg' | 'none'
+  hasRange: false,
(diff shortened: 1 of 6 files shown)


blocks/canvas/editor-utils/selection-toolbar-controller.js
```diff
@@ -0,0 +1,86 @@
...
+  hasRange: false,
+  interacting: false,
+  isNonTextSelection: false,
+  view: null,
+  iframe: null,
+};
+
+function shouldShow() {
+  if (state.mode === 'none') return false;
+  if (state.interacting) return false;
```
sharanyavinod, line 21: Do we need to explicitly hide the toolbar when related interactions are happening? Why not still have the toolbar in the background when the table or other modals are open? We then dont have to worry about rendering the toolbar again on closure.

blocks/canvas/ew-selection-toolbar/ew-selection-toolbar.js
```diff
@@ -272,24 +287,29 @@ class EwSelectionToolbar extends LitElement {
 
   render() {
     const disabled = !this.view;
+    // In wysiwyg mode the iframe owns block-level structure (paragraph/heading,
+    // lists, tables) — only show inline marks and link controls.
+    const inlineOnly = this._mode === 'wysiwyg';
```
sharanyavinod, line 292: In wyswyg, shouldn't the user still be able to change the heading types etc?

sharanyavinod:
In wyswyg, toolbar doesnt appear when I focus at the end of the text intermittently.
Note: Here the cursor is after _Get your cup today!_
Screenshot 2026-06-22 at 10 21 36

PR #1020, opened 2026-06-22
[EW] Improve block library, templates and files

hannessolo:
https://i45674--da-live--adobe.aem.live?nx=dialog-css-vars

https://github.com/adobe/da-nx/pull/523

### DIFF
--- a/blocks/canvas/ew-block-library-modal/ew-block-library-modal.js
+++ b/blocks/canvas/ew-block-library-modal/ew-block-library-modal.js
@@ -15,13 +15,25 @@ const style = await loadStyle(import.meta.url);
 
 const CHEVRON_ICON_SRC = '/img/icons/s2-icon-chevronright-20-n.svg';
 const SEARCH_ICON_SRC = '/img/icons/s2-icon-search-20-n.svg';
+const INFO_ICON_SRC = '/img/icons/s2-icon-infocircle-20-n.svg';
 
-function matchesQuery(text, query) {
-  return (text || '').toLowerCase().includes(query);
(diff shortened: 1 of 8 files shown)


sharanyavinod:
Do you think we should have the add options next to the items as in the da edit page and remove the add/cancel at the bottom? Since the preview stays the same across variations - ie I select Hero/Hero with Text/other variation and preview stays the same throughout, it feels a bit confusing as to what is getting added. User might focus only on the preview, and it might have _default_ displayed, but something else selected on the left. Wdyt? @hannessolo 

Screenshot 2026-06-22 at 20 28 02

blocks/canvas/ew-block-library-modal/ew-block-library-modal.js
```diff
@@ -0,0 +1,364 @@
...
+      <iframe class="modal-preview-frame ${hideIframe}" src=${url}
+              title="Preview of ${name}"
+              allow="clipboard-write *"></iframe>`;
+  }
+
+  render() {
+    const canAdd = !!this._selectedPath
+      && (this._variantsByPath.get(this._selectedPath)?.length || 0) > 0;
+    return html`
+      <dialog class="modal" @close=${this._onDialogClose}>
```
mhaack: Should we use nx-dialog for the dialog

hannessolo: nx-dialog is limited to 480px width. If that's not something we want to stay that way, I can change it...

mhaack (approved):

PR #1048, opened 2026-06-24
fix: count folder contents for delete on Helix 6 (#1034)

### LINKED ISSUE #1034: [hlx6] Delete folder from the browse UI doesn't work
**This issue applies only to running with the Helix 6 backend.**

Just create a folder with a document in it. 
Then try to delete that folder. This fails and you'll get the following dialog confusingly asking to confirm the deletion of 0 items.

Image

chrischrischris:
## Problem

Fixes #1034.

On a **Helix 6** site, deleting a folder from the browse UI shows a confirmation dialog asking to confirm the deletion of **0 items** (and the flow is confusing/broken).

## Root cause

`handleDelete` counts the files inside a selected folder by crawling it. That crawl used nx's `crawl` helper (`tree.js`), which lists **only against the DA origin** (`${DA_ORIGIN}/list`). On a Helix 6 site the content lives behind a different backend, so the crawl found nothing → `_deleteCount` came back `0`.

Every other listing in `da-list.js` already uses `source.list`, which is backend-aware (`isHlx6()` routes to the correct origin for both the legacy DA and Helix 6 backends).

## Fix

- **`blocks/browse/da-list/helpers/utils.js`** — add `crawlDeleteCount`, a recursive file-counter built on the backend-aware `source.list`. It paginates via the continuation token, recurses into subfolders with bounded concurrency, and exposes the same `{ results, cancelCrawl }` shape the old crawl had.
- **`blocks/browse/da-list/da-list.js`** — `handleDelete` now calls `crawlDeleteCount` with the nx2 `source` instead of nx's DA-only `crawl`. Cancellation (`cancelCrawl`) and the mid-crawl bail-out checks are preserved unchanged.
- **`test/unit/blocks/browse/helpers/helpers.test.js`** — new tests: recursive counting, standalone-file inclusion, pagination, cancellation stops recursion, and graceful handling of a non-ok listing.

The actual deletion already worked (it uses backend-aware `source.delete`/`source.move`) — only the confirmation count was broken. This now works for both backends.

## Testing

- Helper unit tests: 16 passed (5 new)
- `da-list` unit tests: 99 passed
- `eslint` on changed files: clean

### DIFF
--- a/blocks/browse/da-list/da-list.js
+++ b/blocks/browse/da-list/da-list.js
@@ -443,17 +443,22 @@ export default class DaList extends LitElement {
 
     this._deleteCountLoading = true;
     try {
-      const { crawl } = await import(`${getNx()}/public/utils/tree.js`);
-      const crawlInstance = crawl({
-        path: folders.map((folder) => folder.path),
-        files,
+      const { source } = await getNx2Api();
+      const { crawlDeleteCount } = await import('./helpers/utils.js');
+      // Use the backend-aware source.list (works for both DA and Helix 6) rather
+      // than nx's crawl, which only lists against the DA origin and returns 0 for
+      // Helix 6 sites.
+      const crawlInstance = crawlDeleteCount({
+        folders,
+        fileCount: files.length,
+        source,
         concurrent: 5,
       });
       this._deleteCrawl = crawlInstance;
-      const allFiles = await crawlInstance.results;
+      const count = await crawlInstance.results;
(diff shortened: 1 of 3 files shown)


chrischrischris:
Closing in favor of the proper fix at the shared layer: adobe/da-nx#555 makes nx2's `crawl` backend-aware via `source.list`, so da-live needs no change — its existing call to nx `crawl` will work on Helix 6 once that ships. This avoids reimplementing folder traversal in da-live just for the delete count.

PR #1071, opened 2026-07-02
fix(ew): don't highlight a folder sharing a name with a file in the files panel

usman-khalid:
Before:
image

After:
image


mhaack:
Good catch. Thanks

mhaack (approved):

PR #1073, opened 2026-07-06
feat(ew): comments

usman-khalid:
To be merged with https://github.com/adobe/da-nx/pull/562

## Summary

Collaborative comments for Experience Workspace.

Users can select text, an image or a table (block) and attach a comment. Each comment is stored as a .json file in `.da/comments/{docId}/{commentId}.json`

When a selection is made, a comment icon is also added to the toolbar in EW to initiate a thread.

Supports replies/threads, resolve/unresolve & orphan threads (user comments on something, the content is deleted while the comment is still active). 

Comment data will survive if a page is moved to another folder.

### Keyboard Shortcuts
- `CMD+Option+M` — if there is a selection, will start composing a new thread.
- `CMD+Enter` to submit a comment/reply

### Read-Only Users
Read only users who should have access to comment only would need to be given write permissions to `/{site}/.da/comments/ + **` in the DA config.

### Layout Mode
~~Layout mode currently doesn't show a toolbar at all when a block is selected. The comment button should be made available there as part of the work done in [SITES-46128](https://jira.corp.adobe.com/browse/SITES-46128)~~

Blocks now have a toolbar since this PR was originally raised. This is now supported.

## How Has This Been Tested?

- Unit tests
- localhost
- https://comments--da-live--adobe.aem.page/canvas?nx=ew-comments&quick-edit=ew-comments&da-admin=stage&da-collab=stage#/usman-khalid/stage-site-2/annotations

## Screenshots (if appropriate)
image

image

## Types of changes

## Checklist


sharanyavinod:
Thanks for your contribution and for your patience!
One architecture note: since you opened this PR, we've standardized how components talk to each other. Communication within the canvas itself now goes through a shared pub/sub bus, and communication that crosses from canvas into the panel has a standard eventing mechanism it should be going through consistently. Could you please update your PR to use these?

usman-khalid:
> Thanks for your contribution and for your patience! One architecture note: since you opened this PR, we've standardized how components talk to each other. Communication within the canvas itself now goes through a shared pub/sub bus, and communication that crosses from canvas into the panel has a standard eventing mechanism it should be going through consistently. Could you please update your PR to use these?

done. routed canvas stuff through canvasBus and canvas > panel through `PANEL_EVENT`

blocks/canvas/ew-tool-panel/tool-panel.js
```diff
@@ -89,6 +112,14 @@ class EwToolPanel extends LitElement {
   }
 
   async _onViewsChange() {
+    // A caller can set `pendingView` (a plain field, not reactive) right before
+    // assigning `views` to request a specific initial view. Consuming it here
+    // lets us skip resolveInitialToolPanelView's flags fetch — which otherwise
+    // races an explicit showPanel(name) call made by that same caller once this
+    // (unawaited) update settles, and can clobber it if the fetch is slow.
```
sharanyavinod: Let's cleanup all the heavy comments - if really essential, lets have a concise, minimal comment instead.

blocks/canvas/ew-comments/iframe-bridge.js
```diff
@@ -0,0 +1,76 @@
...
+      marker.imageSrc = imageSrcAtAnchor(view, range.from);
+    }
+    markers.push(marker);
+  });
+  return markers;
+}
+
+export function postCommentMarkers(port, markers, controller) {
+  port?.postMessage({
+    type: 'set-comment-markers',
```
sharanyavinod: Import and resuse the message-types constants here instead of the string as is. Same for `scroll-to-pos` below.

blocks/canvas/canvas.js
```diff
@@ -19,6 +20,14 @@ import { SEL_BLOCK, SEL_ITEM, SEL_TEXT } from './ew-editor-doc/utils/selection.j
 import { getChatPanelContent } from '../shared/chat-panel.js';
 import { canvasBus } from './utils/canvas-bus.js';
 
+export function handleCommentShortcut(event, { controller, openPanel }) {
```
sharanyavinod: There is a preexisting mechanism for handling keydown + shortcuts in this file. Could you please check if that could be extended instead?

blocks/canvas/comments/comments-panel.css
```diff
@@ -0,0 +1,495 @@
...
+  height: 10px;
+  margin-left: 4px;
+  margin-top: 2px;
+  border: 2px solid currentcolor;
+  border-top-color: transparent;
+  border-radius: 50%;
+  animation: ew-comments-spin 0.6s linear infinite;
+}
+
+@keyframes ew-comments-spin {
```
sharanyavinod: Seems to be duplicated

blocks/canvas/ew-editor-doc/ew-editor-doc.js
```diff
@@ -43,6 +51,65 @@ const { CHAT_EVENT } = await import(`${getNx()}/utils/chat.js`);
...
 const style = await loadStyle(import.meta.url);
+const commentHighlightStyle = await getSheet('/blocks/canvas/comments/comment-highlight.css');
+
+export function createCommentsStoreFor(session, ctx) {
+  return session?.docId
+    ? createCommentsStore({ docId: session.docId, owner: ctx.org, repo: ctx.repo })
+    : null;
+}
+
+export function publishCommentsController(store, wsProvider) {
```
sharanyavinod: Looks to be unused?

blocks/canvas/ew-panel-extensions/helpers.js
```diff
@@ -541,6 +543,62 @@ function createVersioningView() {
...
+    },
+    load: async () => {
+      await import('../comments/comments-panel.js');
+      const el = document.createElement('ew-comments');
+      // The tool panel owns the close control in EW, so hide the in-panel one.
+      el.embedded = true;
+      const getController = () => getCommentsBridge().controller;
+      el.controller = getController();
+      const syncPanelOpen = bindPanelOpenToVisibility(el, getController);
+      canvasBus.commentsControllerState.subscribe((controller) => {
```
sharanyavinod: Would be good to handle unsubscribe on panel close

blocks/canvas/comments/helpers/templates.js
```diff
@@ -0,0 +1,295 @@
...
+    <div class="ew-comments-list">
+      <p class="ew-comments-hint">
+        Select content and press <kbd>${COMMENT_SHORTCUT}</kbd> to add a comment.
+      </p>
+      ${tabs.length > 1 ? html`
+        <div class="ew-comment-tabs" role="tablist" aria-label="Comment thread categories">
+          ${tabs.map((tab) => html`
+            <button
+              class="ew-comment-tab ${panel._activeTab === tab.id ? 'is-active' : ''}"
+              role="tab"
```
sharanyavinod: role tabpanel seems to be missing here - please check if it can be included and if not, tablist/tab might not be the best option here

blocks/canvas/comments/comments-panel.css
```diff
@@ -0,0 +1,495 @@
...
+
+.ew-comments-title {
+  top: 0;
+  background: var(--comment-surface);
+  padding-top: 12px;
+  padding-bottom: 12px;
+  z-index: 10;
+}
+
+.ew-comments-close-btn {
```
sharanyavinod: Wherever possible, please reuse the styles available in https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html

blocks/canvas/ew-editor-doc/ew-editor-doc.js
```diff
@@ -284,49 +366,87 @@ export class EwEditorDoc extends LitElement {
...
+        } else {
+          const onSynced = (isSynced) => {
+            if (!isSynced) return;
+            wsProvider.off('synced', onSynced);
+            doLoad();
+          };
+          wsProvider.on('synced', onSynced);
+        }
+      }
+
```
sharanyavinod, line 358: Quite a lot of comment specific logic is introduced here - ideally we want to keep this file to have only core doc editing logic. Would be nice to explore how to extract these and call in a clean way instead. If not all, as much as possible.

blocks/canvas/canvas.js
```diff
@@ -245,6 +263,13 @@ export default async function decorate(block) {
 
   const header = await installCanvasHeader(block, { org, site });
 
+  window.addEventListener('keydown', (event) => {
+    handleCommentShortcut(event, {
+      controller: getCommentsBridge().controller,
+      openPanel: () => openPanelSection('tools', 'comments'),
```
sharanyavinod: This could be done within the existing shortcut handler as well -> version creation does something similar. Could you please check if the same approach might work here?

blocks/canvas/ew-panel-extensions/helpers.js
```diff
@@ -541,6 +543,62 @@ function createVersioningView() {
...
+ * while the panel is visible: tie the controller's panelOpen flag to whether the
+ * panel element is actually rendered (the active tool-panel view in an open
+ * rail). Returns a function that re-applies the current visibility to whatever
+ * controller `getController` currently resolves, for use after a controller swap.
+ */
+export function bindPanelOpenToVisibility(el, getController) {
+  let visible = false;
+  let hideTimer = null;
+  const syncPanelOpen = () => getController()?.setPanelOpen(visible);
+  const observer = new IntersectionObserver((entries) => {
```
sharanyavinod: I think intersection observer might be slightly overkill here - tool-panel already knows which is the currently active panel. Couldnt we expose this so all who want to know which panel is active only refer to it?

blocks/canvas/comments/helpers/templates.js
```diff
@@ -0,0 +1,295 @@
...
+          <span class="ew-comment-author">${comment.author.name}</span>
+          ${isDetached ? renderDetachedReference(comment) : nothing}
+          <span class="ew-comment-time" title="${formatUtils.formatFullTimestamp(comment.createdAt)}">
+            ${formatUtils.formatTimestamp(comment.createdAt)}
+          </span>
+        </div>
+        ${showResolve || showMenu ? html`
+          <div class="ew-comment-header-actions" @click=${(e) => e.stopPropagation()}>
+            ${showResolve ? html`
+              <button class="ew-comments-btn-resolve" ?disabled=${!!panel._submittingId} @click=${() => panel.handleResolveThread(threadId)} title="Resolve">
```
sharanyavinod: aria-label is missing here

blocks/canvas/canvas.js
usman-khalid: Thanks, this was leftover from when this was done for DA and there the shortcut mechanism was gated for read-only users. Updated to use existing for canvas.

blocks/canvas/comments/helpers/templates.js
usman-khalid: Changed to `role="group"`

blocks/canvas/comments/comments-panel.css
usman-khalid: - Removed this
- Used nx action button where possible (thread menu and resolve buttons)
- used nx form buttons for the commenting form

blocks/canvas/ew-editor-doc/ew-editor-doc.js
usman-khalid: Extracted this out into `utils/editor-comments.js`

blocks/canvas/ew-panel-extensions/helpers.js
usman-khalid: we don't show the comment markers in the canvas if the rail is closed entirely, which is what this is for. doing the above would  cause the them to still be visible if a user had the comments panel open, then closed the right rail entirely.

sharanyavinod: There is already an event which is fired when the panel is closed. Couldn't we listen to that instead, and thus wouldn't a combination of this + what is the panel currently active cover this case without the need for an observer?

sharanyavinod (dismissed):

blocks/canvas/comments/helpers/controller.js
```diff
@@ -0,0 +1,311 @@
...
+
+    async deleteComment({ commentId }) {
+      const comment = store.get(commentId);
+      if (!comment) return;
+      if (comment.threadId == null) {
+        const replyIds = [];
+        store.forEach((entry, id) => {
+          if (entry.threadId === commentId) replyIds.push(id);
+        });
+        await store.deleteBatch([...replyIds, commentId]);
```
mhaack: This is not atomic, if one delete fails it can leave orphaned entries.

usman-khalid: good call - each comment is a separate json, so technically the root thread comment could win and leave the rest as orphans here as well, and that would make the replies unavailable in the UI. updated this to `deleteInOrder` which aborts on the first failure and deletes the root comment last, so the worst case really is a thread without n number of replies, which the user can retry from the UI.

mhaack (approved):

siwong4:
Hi, does the comments section support tagging users to notify them when their attention is required.

PR #1074, opened 2026-07-06
fix(sheet): persist row/column delete and move in sheet editor

shsteimer:
## What changed
`ondeleterow`, `ondeletecolumn`, `onmoverow`, and `onmovecolumn` are now wired to trigger a save, the same as `onafterchanges` already does for cell edits. Row/column inserts are left unhooked - a fresh row/column starts empty, and once real content lands in it, `onafterchanges` captures it anyway.

Also adds Playwright e2e tests that drive a real jspreadsheet-ce grid end-to-end (delete two rows, delete two columns, reload, verify the deletion persisted), plus a unit test following this repo's existing `handleSave` testing convention, covering all four new hooks.

Row/column move (`onmoverow`/`onmovecolumn`) are covered at the unit level only - e2e coverage would need to simulate real drag-and-drop, which is more fragile than a context-menu click for comparatively low added value given the unit tests + library source already confirm the mutation path is symmetric to delete.

## Why
jspreadsheet-ce only fires `onafterchanges` for cell-value edits. Deleting or reordering rows/columns never marked the sheet dirty or triggered a save, so those changes were silently lost on reload unless followed by an unrelated cell edit or an explicit Save click.

### DIFF
--- a/blocks/sheet/utils/index.js
+++ b/blocks/sheet/utils/index.js
@@ -26,11 +26,11 @@ function finishSetup(el, data) {
     const save = () => handleSave(el.jexcel, el.details.view);
     sheet.options.onafterchanges = save;
     // onafterchanges doesn't fire for these; inserts are skipped since they start empty
-    // and get saved once onafterchanges fires on their content.
+    // and get saved once onafterchanges fires on their content. onmovecolumn is left
+    // unbound - columnDrag is off, so there's no UI path that can ever fire it.
     sheet.options.ondeleterow = save;
     sheet.options.ondeletecolumn = save;
     sheet.options.onmoverow = save;
-    sheet.options.onmovecolumn = save;
   });
 
   // Setup tabs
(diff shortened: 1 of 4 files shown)


blocks/sheet/utils/index.js
```diff
@@ -23,9 +23,15 @@ function finishSetup(el, data) {
...
-    sheet.options.onafterchanges = () => {
-      handleSave(el.jexcel, el.details.view);
-    };
+    const save = () => handleSave(el.jexcel, el.details.view);
+    sheet.options.onafterchanges = save;
+    // Removing or reordering existing rows/columns doesn't fire onafterchanges, so those
+    // need their own hooks or the change is silently lost. Inserts are skipped - a fresh
+    // row/column starts empty, and once data lands in it onafterchanges saves it anyway.
+    // onsort is skipped too: it's a header-click view convenience, not an intentional edit.
+    sheet.options.ondeleterow = save;
```
shsteimer, line 31: genuinely wans't sure if we should include sort here or not. to me, sort is a user level operation to see data in a different way, but I could certainly see the case for wanting sort to be persisted.

hannessolo (commented):
I think not saving on sort is fine

hannessolo (dismissed):

hannessolo (approved):

PR #1077, opened 2026-07-07
fix(e2e): reduce flakiness in Update Document and collab cursor tests

shsteimer:
## What changed
Two pre-existing e2e tests showed up as flaky in CI while working on #1074 (unrelated PR) - "Update Document" and "Collab cursors in multiple editors". Root-caused and fixed both.

### Update Document
Raced a fixed `waitForTimeout(3000)` against Y.js persistence before closing the page and reopening it in a new one. Occasionally too short under CI load, so the reopened page loaded an empty document. Replaced with `da-content.forceSave()` - the same flush-and-wait-for-server-ack protocol the app already uses before preview/publish - so the test deterministically confirms the save landed instead of guessing at a timeout.

### Collab cursors in multiple editors
`span.ProseMirror-yjs-cursor` visibility depends on a real Y.js awareness broadcast round trip between two browser tabs/websocket connections. That occasionally exceeded the global 15s `expect` timeout under slower CI runners (seen failing on firefox/webkit, not chromium). Gave those specific assertions (cloud icon + cursor visibility) an explicit 30s timeout and bumped the test-level timeout from 60s to 120s to match - no code path bug found, this is inherent network/CI timing variance.

## Why
Found while debugging unrelated CI failures on #1074 and wanted to fix separately rather than bundle into that PR.

### DIFF
--- a/test/e2e/tests/authenticated/collab.spec.js
+++ b/test/e2e/tests/authenticated/collab.spec.js
@@ -17,7 +17,7 @@ test('Collab cursors in multiple editors', async ({ browser, page, browserName }
   // Ensure that the edits are visible to both and that the collab cursors are there
   // Also check that the cloud icon is visible for the collaborator
 
-  test.setTimeout(60000);
+  test.setTimeout(120000);
 
   const pageURL = getTestPageURL('collab', workerInfo);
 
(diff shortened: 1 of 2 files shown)


shsteimer:
Closing for now, will reopen later.

PR #1084, opened 2026-07-08
feat(ew): add support for blocks search in slash menu

usman-khalid:
Adds support to slash menu to pick blocks from the library if configured.

Additionally:
- Dont' show "Open Block Library" if no library is configured
- Show a loading state for the library if the user starts typing before the library is fetched in the slash menu.

Test URL: https://ewsta--da-live--adobe.aem.page/canvas?da-admin=stage&da-collab=stage#/usman-khalid/stage-site-2/comments

Should be merged with: https://github.com/adobe/da-nx/pull/585

### DIFF
--- a/blocks/canvas/editor-utils/block-slash.js
+++ b/blocks/canvas/editor-utils/block-slash.js
@@ -102,9 +102,8 @@ export async function prefetchBlockLibrary({ org, site } = {}) {
   store.key = key;
   store.state = 'loading';
   try {
-    const { fetchExtensions, fetchBlocks } = await import('../ew-panel-extensions/helpers.js');
-    const extensions = await fetchExtensions(org, site);
-    const ext = extensions?.find((e) => e.name === 'blocks');
+    const { loadBlockLibrary } = await import('../ew-panel-extensions/helpers.js');
+    const { ext, blocks } = await loadBlockLibrary(org, site);
     if (!ext) {
       store.entries = [];
       store.hasLibrary = false;
@@ -113,7 +112,6 @@ export async function prefetchBlockLibrary({ org, site } = {}) {
     }
 
     store.hasLibrary = true;
-    const blocks = await fetchBlocks(ext.sources);
     await ingestBlocks(blocks);
   } catch {
     store.state = 'empty';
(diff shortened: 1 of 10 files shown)


blocks/canvas/ew-editor-doc/prose.js
```diff
@@ -125,6 +149,7 @@ export default async function initProse({
   });
 
   addSyncedListener(wsProvider, canWrite, setEditable);
+  prefetchBlocksAfterSync(wsProvider, canWrite);
```
hannessolo: I wonder if we could wait until the user types `/` to fetch this? Probably there would be enough time int he time it takes to type the block name?

blocks/canvas/editor-utils/block-slash.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export async function insertBlockItem(view, id) {
+  const entry = store.entries.find((e) => e.id === id);
+  if (!entry) return;
+  const { insertBlock } = await import('../ew-panel-extensions/helpers.js');
+  insertBlock(view, entry.dom);
+}
+
+export async function prefetchBlockLibrary({ org, site } = {}) {
```
hannessolo: Could we reuse this function between here and the block library?

hannessolo (commented):
Amazing feature!

blocks/canvas/ew-editor-doc/prose.js
usman-khalid: I thought about this - but IMO power users are pretty quick if they know what they want from the slash menu. Currently prefetches after the doc is loaded which should be fine? @mhaack thoughts?

blocks/canvas/editor-utils/block-slash.js
usman-khalid: Done.

blocks/canvas/ew-editor-doc/prose.js
mhaack: If we wait until the user types `/` and the user types fast like `/h` would the only show the heading options or also the hero block.

I would also prefer to load it only if user type`/`. There are huge block libraries in some projects like a.com

usman-khalid: Ok - done. Library will be fetched when the slash menu is invoked now. 

If the user quickly types something before it's done loading, we show a loading indicator.

mhaack (approved):

PR #1086, opened 2026-07-09
fix: surface image upload failures instead of failing silently

kptdobe:
## Summary
- A failed image upload (`imageDrop.js`) silently returned, leaving the FPO placeholder stuck in the doc forever with zero feedback.
- Now: the placeholder is removed, the failure is logged, and the existing `da-title` error banner (same UI used for save/publish failures) shows the filename and server status.

## Test plan
- [x] Added a failing-first test asserting the placeholder is removed, the error is logged, and `da-title._status` is set — confirmed red before the fix
- [x] Full prose plugin suite: 299/299 passing
- [x] `npm run lint` — clean


mhaack (approved):
@hannessolo we should adapt this for EW as well

PR #1088, opened 2026-07-10
fix(ew): exclude section-metadata from block indexing to fix outline panel offset

usman-khalid:
Fixes a bug in the EW page outline where blocks after section-metadata were selected incorrectly, because `getBlockPositions` counted section-metadata tables while the outline strips them.

Fix is to skip section-metadata (and page metadata) in `getBlockPositions` so editor indices match the outline

BEFORE:

AFTER:


mhaack (approved):
LGTM

PR #1092, opened 2026-07-13
fix(hlx6): correct paste destination path for hlx6 sites in browse

### LINKED ISSUE #1032: [hlx6] copy+paste document in the same directory doesn't work
**This issue applies only to running with the Helix 6 backend.**

Take a document in a folder, that has some content in it. 

Copy it and paste it back into the same folder.

A new document appears with the name `doc-copy`. Open it and see that it doesn't exist yet.

On further inspection, the document seems to have been copied to the wrong place, as a new folder /org/site/org/site/folder/doc-copy has appeared (note the duplication of org/site).

See this video for a demonstration:

https://github.com/user-attachments/assets/2785bfd3-e88f-4cf8-9006-479d17ed7980

### LINKED ISSUE #1033: [hlx6] copy+paste document across directories doesn't work
**This issue applies only to running with the Helix 6 backend.**

Copy and paste of a document across folders doesn't work. Similar to #1032 but then across folders. 

Filed this as a separate issue to ensure that it gets validated too.

mhaack:
fixes: #1032
fixes: #1033

merge after: https://github.com/adobe/da-nx/pull/594 + re-run playwright tests

## Problem
In `da-list.js` `handlePaste`, the destination path for pasted items on hlx6 sites was computed incorrectly:
- non-hlx6: destination should be `/org/site/path/filename`
- hlx6: destination should be `/path/filename` (no org/site segments)

## Fix
- After building the paste item's destination, if `this._isHlx6` is true, strip the `org`/`site` segments via `sanitizePathParts` so the destination is folder-relative.
- Also fixes `handleItemAction`'s folder detection to use `isFolder` (checks `item.ext` or `item.contentType === 'application/folder'`) consistently for choosing `copyFolder`/`deleteFolder` vs single-item `copy`/`delete`.

## Testing
- Ran eslint on the changed file, no issues.

Test:
- https://hlx6fc--da-live--adobe.aem.live/#/da-testautomation/da-e2e-tests/blah/mh-test

### DIFF
--- a/blocks/browse/da-list/da-list.js
+++ b/blocks/browse/da-list/da-list.js
@@ -363,9 +363,12 @@ export default class DaList extends LitElement {
   async handleItemAction({ item, type = 'copy' }) {
     const { source } = await getNx2Api();
 
-    const useDeleteFolder = type === 'delete' && !item.ext;
-    const deleteFn = useDeleteFolder ? source.deleteFolder : source.delete;
-    const type2fn = { copy: source.copy, delete: deleteFn, move: source.move };
+    const isFolder = !item.ext || item.contentType === 'application/folder';
+    const type2fn = {
+      copy: isFolder ? source.copyFolder : source.copy,
+      delete: isFolder ? source.deleteFolder : source.delete,
+      move: source.move,
+    };
     const fn = type2fn[type];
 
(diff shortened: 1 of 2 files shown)


mhaack:
While the copy issues is fixed with this PR the e2e tests still fail with page rename. Looking into it.

PR #1104, opened 2026-07-14
fix(shared): default DA admin to stage on localhost

benpeter:
### Problem

Opening a site in the local dev server (e.g. `http://localhost:3020/#/benpeter/hlx6-test`) redirects to `/not-found`, even though the same site loads fine on prod da.live.

Root cause is an environment mismatch between da-live and da-nx for the `localhost` host:

- On `localhost`, the IMS token is **always minted against IMS stage** (da-nx's `env` resolver maps localhost -> `dev`, and `dev` -> the `stg1` IMS environment). This is not overridable per request.
- da-live's own env resolver (`blocks/shared/constants.js` `getDaEnv`) had no concept of localhost and defaulted every backend to **prod**. So `DA_ORIGIN` pointed at `admin.da.live` (prod).
- A stage IMS token sent to prod admin gets a **401**. da-live's `daFetch` (`blocks/shared/utils.js`) redirects to `/not-found` on a 401 against a DA origin after one refresh attempt — hence the redirect.

da-nx (nx2) already defaults its `DA_ADMIN` to stage on localhost, so the two halves of the app disagreed about which admin to talk to.

### Fix

Align da-live's admin default with da-nx: on localhost, `DA_ORIGIN` / `getDaAdmin()` now default to **stage-admin.da.live**, matching the stage IMS token. A stage token now reaches stage admin, which returns 403/404 (not 401) for a not-yet-authorized config read — and da-live's `daFetch` returns on 403 rather than redirecting, so the browser falls through to the default editor instead of `/not-found`.

Scope is deliberately narrow — **admin only**. Content, collab, and preview origins keep their existing prod default (with the existing `?da-content=stage` / `?da-collab=stage` manual overrides unchanged). Explicit overrides (`?da-admin=prod`, `?da-admin=reset`) still win everywhere.

`getDaEnv` gains an optional `localDefault` parameter (defaults to `prod`, so all other callers are unchanged) and now keys off `url.hostname.includes("local")`, matching da-nx's own localhost detection.

### Verification

- Unit suite green (1603 passed, 4 skipped); lint clean.
- New tests in `test/unit/blocks/shared/constants.test.js` assert localhost -> stage-admin, non-local -> prod, and explicit-override precedence.
- Live-tested against the running dev server: the served `constants.js` resolves `DA_ORIGIN = https://stage-admin.da.live` on localhost, `https://admin.da.live` with `?da-admin=prod`, and `https://admin.da.live` on prod da.live; content/collab origins unchanged in all cases.

### Follow-up

da-live maintaining its own env resolver in parallel with da-nx's is the underlying source of this drift. A follow-up could consolidate da-live onto nx2's `DA_ADMIN` / env resolution so the two can never disagree again.


blocks/shared/constants.js
```diff
@@ -59,12 +64,12 @@ export const getDaAdmin = (() => {
   let daAdmin;
   return (location) => {
     if (!location && daAdmin) return daAdmin;
-    daAdmin = getDaEnv(location || window.location, 'da-admin', DA_ADMIN_ENVS);
+    daAdmin = getDaEnv(location || window.location, 'da-admin', DA_ADMIN_ENVS, 'stage');
```
kptdobe: I do not like that `stage` becomes the default. Default must be `prod`, nothing else.

blocks/shared/constants.js
```diff
@@ -42,15 +42,20 @@ const DA_ETC_ENVS = {
...
+  const query = url.searchParams.get(key);
   if (query && query === 'reset') {
     localStorage.removeItem(key);
   } else if (query) {
     localStorage.setItem(key, query);
   }
-  const env = envs[localStorage.getItem(key) || 'prod'];
+  // On localhost the IMS token is always minted against IMS stage, so the DA
+  // admin default must be stage there too — otherwise a stage token hits prod
+  // admin, 401s, and da-nx redirects to /not-found.
```
kptdobe: `localhost` is the dev env, it should be as invisible as possible in the code which should be prod ready. Of course, we do not have the choice to have code for some "local dev" scenario but they should be silent. Just remove the comments.

blocks/shared/constants.js
```diff
@@ -59,12 +64,12 @@ export const getDaAdmin = (() => {
...
   return (location) => {
     if (!location && daAdmin) return daAdmin;
-    daAdmin = getDaEnv(location || window.location, 'da-admin', DA_ADMIN_ENVS);
+    daAdmin = getDaEnv(location || window.location, 'da-admin', DA_ADMIN_ENVS, 'stage');
     return daAdmin;
   };
 })();
 
-export const DA_ORIGIN = (() => getDaEnv(window.location, 'da-admin', DA_ADMIN_ENVS))();
+export const DA_ORIGIN = (() => getDaEnv(window.location, 'da-admin', DA_ADMIN_ENVS, 'stage'))();
```
kptdobe: Same here.

kptdobe (changes requested):

blocks/shared/constants.js
benpeter: defaulting to prod throughout now (it was already, technically, but didn't read like that). I kept `stage` for localhost, as defaulting to prod would break local dev unless someone knows to override with `?da-admin=stage` (or `?da-admin=local` if they're running the admin worker).

PR #1108, opened 2026-07-14
fix(sheet): save active cell before leaving sheet

### LINKED ISSUE #1097: [hlx6] sheets are not saved correct
To reproduce
* Create an empty sheet.
* Add a header row.
* Add a first line.
* Go back in to browse
* Re-open sheet
* > last cell data is missing

benpeter:
## Summary

Commit the active spreadsheet cell and flush the pending autosave before navigating to the parent folder. Queued saves retain the sheet path where the edit occurred.

Fixes #1097.

## Live test

Edited the final cell, immediately navigated to the parent folder without tabbing out or waiting for the debounce, then reopened the sheet. The edited value was retained.

### DIFF
--- a/blocks/edit/da-title/da-title.js
+++ b/blocks/edit/da-title/da-title.js
@@ -215,6 +215,14 @@ export default class DaTitle extends LitElement {
     }
   }
 
+  async handleNavigate(e) {
+    const modified = e.button !== 0 || e.metaKey || e.ctrlKey || e.shiftKey || e.altKey;
+    if (!this.beforeNavigate || modified) return;
+    e.preventDefault();
+    const { href } = e.currentTarget;
+    if (await this.beforeNavigate() !== false) window.location.href = href;
+  }
+
   async handleAction(action) {
     this._status = null;
     this._isSending = true;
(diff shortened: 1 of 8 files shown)


hannessolo:
Thanks a lot for your PR.

I see the issues this is trying to fix but I'm not yet convinced this is the right fix:
- When the user navigates away using any other method (close the tab, close the browser, use the browser back button, ...) this won't catch it
- I personally consider the behaviour that a cell you didn't focus out of isn't saved intentional - if you hit `esc` while editing the cell, the edit is undone, so in my mind it isn't finalised until you hit enter or focus out. This is consistent with how google sheets behaves, although that warns you if you try to close the tab with unsaved changes

That said I'm open to be convinced otherwise - @mhaack @shsteimer what do you think?

hannessolo:
Something we were discussing to add regardless is an indicator to show whether the page has been saved.

benpeter:
thanks @hannessolo - I can relate, expectations might vary a lot and probably indicator + a warning are the right middle ground.

also, nothing new as noted on #1097, but maybe @mhaack saw something that I didn't

benpeter:
webkit tests fail, so I'll push a fix for that either way so this PR is in consistent state

hannessolo:
Can you take a look at https://github.com/adobe/da-live/pull/1150 as an alternative?

benpeter:
thanks @hannessolo -- closing in favor of #1150

PR #1113, opened 2026-07-15
fix(sheet): serialize reads behind in-flight saves to prevent drift m…

hannessolo:
Draft - still buggy

https://sheetseq--da-live--adobe.aem.live

This PR fixes a bug where rapidly typing in the sheet editor either loses updates or causes the "out of sync" popup to appear.

Approach:

- Switch drift detection from full-body JSON comparison to ETag comparison
- Add `_pendingSave` promise so at most one POST is in flight
- Gate `checkForDrift` and live-file `getData` on `awaitPendingSave` so no read races a same-tab write - since reads are much faster the polling could hit this case
- Move the Save/Preview/Publish path (`da-title.handleAction`) inside `runSave` so it can't race a background autosave

### DIFF
--- a/blocks/edit/da-title/da-title.js
+++ b/blocks/edit/da-title/da-title.js
@@ -226,31 +226,39 @@ export default class DaTitle extends LitElement {
 
     // Bail before writing if the remote drifted under us — protects against
     // last-write-wins. Drift triggers the stale-content dialog via onStale.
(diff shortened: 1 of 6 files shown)


kptdobe (approved):
Tested. Works as expected now.

usman-khalid (approved):

PR #1114, opened 2026-07-15
feat(ew): sync layout-mode block & image selection with the doc

usman-khalid:
Should be merged with https://github.com/adobe/da-nx/pull/605

Implements support to select blocks/images in canvas synced to doc and the outline panel.
- If in split mode, clicking a block in the outline panel scrolls down to it in both doc-mode and canvas.
- Selecting a block in canvas retains the selection in doc-mode (and vice versa)
- Selecting a block in canvas adds it to the chat context, same as doc-mode
- Allows synced image selection between doc-mode and canvas

https://github.com/user-attachments/assets/bb1663b5-fc9b-46cf-b7fa-022a84476b69

**Note on event system consistency:** This PR adds one new postMessage pair (`set-selected-node` / `node-select`) and one new consumer of the existing `editorSelectChange` bus. Both follow current conventions, but I'm aware of [SITES-46045](https://jira.corp.adobe.com/browse/SITES-46045) and if that lands while this is still open, happy to refactor as needed.

### DIFF
--- a/blocks/canvas/editor-utils/editor-utils.js
+++ b/blocks/canvas/editor-utils/editor-utils.js
@@ -192,6 +192,21 @@ export function getInstrumentedHTML(view) {
     div.insertAdjacentElement('beforebegin', blockMarker);
   });
 
+  const originalImages = view.dom.querySelectorAll('img');
+  const clonedImages = editorClone.querySelectorAll('img');
+  originalImages.forEach((originalImage, index) => {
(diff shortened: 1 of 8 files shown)


sharanyavinod (dismissed):
Thank you for the contribution! 

Looks good, just a nit: would be nice to cleanup the comments where possible or make them more concise.

sharanyavinod (approved):

PR #1115, opened 2026-07-15
fix(ew): add Block Library to the tool panel dropdown

amol-anand:
# Adds Block Library to the ew canvas tool panel dropdown

## Description
Adds Block Library to the right tool panel dropdown, alongside the other tools (Outline, Files, Versions, Templates, Icons, Placeholders, AEM Assets).

Selecting it opens the same block library dialog that the slash menu and outline "+" button already open — inserting a variant drops it at the current cursor position in the editor.

## Motivation and Context
Block Library is the odd one out: every other tool is reachable from the tool panel dropdown, but the block library was only accessible via the slash menu and a small, easy-to-miss + button on the outline section header rows. This puts it where users already look for tools, so it's discoverable in the same place as everything else — without removing the existing entry points.

Rather than rendering it inline like the other library tools (its tree + preview layout needs far more room than the side panel offers), the dropdown entry reuses the existing openBlockLibraryModal() flow. This keeps a single block-library UI and avoids offering two different experiences for the same feature.

**Implementation**
`extensionToPanelView` special-cases the blocks extension to produce a view with a new `experience: 'modal'` and an `openModal()` that calls the existing `openBlockLibraryModal()`, wiring insertion into the active editor view.
`ew-tool-panel` handles the new modal experience in `showPanel` — mirroring the existing window case: it fires the action and returns without mounting anything inline or changing the active tab. The dropdown item shows the "opens externally" affordance.

No changes to the block library modal component or its styles — it's the exact same dialog, now reachable from a third entry point.

## How Has This Been Tested?

https://shwblock--da-live--adobe.aem.live/canvas#/aem-sandbox/block-collection/drafts/amol/shwblock

Verified Block Library appears in the tool panel dropdown for a site with a configured blocks library, grouped under Library.
Selecting it opens the block library dialog; searching, expanding blocks, and previewing variants all work as before.
Inserting a variant places the block at the current cursor position in the editor.
Confirmed the existing entry points (slash menu, outline section "+" button) still work unchanged.
Confirmed the dropdown entry does not appear for sites with no blocks library configured.

## Types of changes

## Checklist:

### DIFF
--- a/blocks/canvas/ew-panel-extensions/helpers.js
+++ b/blocks/canvas/ew-panel-extensions/helpers.js
@@ -470,7 +470,7 @@ function createVersioningView() {
   };
 }
 
-function extensionToPanelView(ext, section) {
+export function extensionToPanelView(ext, section) {
   // Block library opens its own dedicated modal (used by the slash menu and
   // outline "+" button) rather than the generic inline panel or iframe dialog.
   if (ext.name === 'blocks') {
--- a/blocks/canvas/ew-panel-extensions/helpers.js
+++ b/blocks/canvas/ew-panel-extensions/helpers.js
@@ -471,6 +471,28 @@ function createVersioningView() {
 }
 
 function extensionToPanelView(ext, section) {
+  // Block library opens its own dedicated modal (used by the slash menu and
+  // outline "+" button) rather than the generic inline panel or iframe dialog.
+  if (ext.name === 'blocks') {
+    return {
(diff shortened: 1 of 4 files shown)


mhaack:
@amol-anand can you have a look at the link issues and we should add some tests as well

mhaack (approved):

sharanyavinod (approved):

PR #1119, opened 2026-07-16
fix: Re-enable version tests for HLX6

mhaack:
## Description

This re-enables the version e2e test, with the exception of the audit trail part that is skipped, as we don't have an audit trail in HLX6


mhaack:
@bosschaert what do you think? Is this fine for now? If yes, feel free to merge.

sharanyavinod (approved):

PR #1120, opened 2026-07-17
refactor: cleanup events

sharanyavinod:
## Description

- remove unnecessary close event in da-link and da-alt dialogs
- use a common constant for event type across boundary ie canvas <-> quickedit
- standardize event structure for postmessage calls with quickedit ie always follow pattern type and payload
- Requires https://github.com/adobe/da-nx/pull/607
- Preview: https://ewevt--da-live--adobe.aem.live/canvas?nx=ewevt&quickedit=ewevt#/exp-workspace/frescopa/coffee

Notes:
- common file within nx and not nx2 since quickedit lives in nx
- both sides support both patterns currently. To deprecate once things are up and running at both ends.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.


scripts/utils.js
```diff
@@ -87,6 +87,13 @@ export const getNx2 = () => {
   return nx.endsWith('/nx') ? `${nx}2` : nx;
 };
 
+// Forces the base nx (never nx2), regardless of the page's nxVer flag — for consumers
+// like da-nx's quick-edit plugin that only exist under nx/, not nx2/.
+export const getNx1 = () => {
```
hannessolo: I'm slightly concerned adding this will be abused (the fix in 99% of cases is porting code over to nx2, not importing from nx1), the only reason we can't do it here is because we don't control all consumers that import quick edit.

blocks/edit/da-preview/da-preview.js
```diff
@@ -59,12 +60,15 @@ export default class DaPreview extends LitElement {
...
+    const targetOrigin = getIframeOrigin(target);
+    if (!targetOrigin) return;
+
     const channel = new MessageChannel();
     this.port1 = channel.port1;
     this.port2 = channel.port2;
 
     // attempt to send a message
-    target.contentWindow.postMessage({ ready: true }, '*', [this.port2]);
+    target.contentWindow.postMessage({ ready: true }, targetOrigin, [this.port2]);
```
hannessolo: IIRC we changed this to * intentionally - I think it was to support the claude design to EW use case. I need to double check if we can remove it again though.

hannessolo: We still need this - here is what claude told me:


> The file/line: blocks/canvas/ew-editor-wysiwyg/ew-editor-wysiwyg.js:169, inside _postQuickEditInitToIframe. Currently:
  iframe.contentWindow.postMessage({ init: config, location }, '*', [port2]);
  This is Karl's exact { init: config, location } WYSIWYG init handshake from the Slack thread.

 > What PR #1120 changes it to: it replaces '*' with a new getIframeOrigin(iframe) helper (added in blocks/shared/utils.js):
```
  export function getIframeOrigin(iframe) {
    try {
      return new URL(iframe.src, window.location.href).origin;
    } catch { return null; }
  }
```

So at least for that path we should preserve *

sharanyavinod: Thank you for checking - I'll revert these and for the rest reuse https://github.com/adobe/da-live/pull/1154 which was merged in the meanwhile

sharanyavinod: Done

hannessolo (approved):

PR #1147, opened 2026-07-17
feat(edit): loading spinner for pages and sheets

markdaugherty:
## What changed
Adds a centered loading spinner to the page editor and the sheet view while their content loads. For pages, the spinner shows until the document has synced over the collab WebSocket and rendered (gated on a new `contentReady` flag set from `onWsSync`), not merely when the empty editor mounts. For sheets, it shows while data is fetched and the grid is rendered. Reuses the existing `.da-loading-spinner` visual style.

## Why
Pages and sheets previously rendered an empty area until content finished loading, with no indication anything was happening.

## Reviewer notes
- The page spinner deliberately persists through the mounted-but-empty-editor phase; it clears on the same `synced` signal already used to enable editing, so content and spinner-removal stay in sync.
- `.da-loading-spinner` / `@keyframes da-spin` are redefined locally in `da-content.css` and `sheet.css` (sized up to 32px) rather than importing `shared/styles/base.css`, since neither view adopts that sheet.
- Unit tests added for both behaviors (da-content spinner render + prose `contentReady` on sync).


mhaack:
@markdaugherty can you have a look at the failing e2e tests

blocks/edit/prose/index.js
```diff
@@ -431,13 +431,15 @@ function restoreCursorPosition(view) {
...
-function addSyncedListener(wsProvider, canWrite) {
+function addSyncedListener(wsProvider, canWrite, daContent) {
   onWsSync(wsProvider, () => {
     if (canWrite) {
       const pm = document.querySelector('da-content')?.shadowRoot
         .querySelector('da-editor')?.shadowRoot.querySelector('.ProseMirror');
       if (pm) pm.contentEditable = 'true';
     }
+    // Content has synced and rendered — clear the loading spinner.
+    if (daContent) daContent.contentReady = true;
```
sharanyavinod, line 442: Shouldn't this be reset on nav as well? Currently the spinner appears only on the very first /edit -> all subsequent pages still have the previous behavior.

PR #1148, opened 2026-07-19
fix(canvas): convert fragment URLs to relative for live preview so that fragment block renders content

### LINKED ISSUE #1126: [EW] EDS fragments are not rendered in layout mode
It just shows the link instead of rendering the fragment. Example (BACOM blog): https://da.live/canvas#/aemsites/da-bacom-blog/drafts/buergi/rikio

Source: https://jira.corp.adobe.com/browse/SITES-47170

kmurugulla:
## Description

- Convert absolute EDS/Helix fragment URLs to relative paths during live preview
- Only applies to links containing `/fragments/` in their pathname
- Minimal blast radius: only affects live preview rendering, not saved HTML

## Related Issue

Fixes #1126

## How Has This Been Tested?

- Added 9 test cases covering fragment URL conversion scenarios
- All tests passing (1658 passed, 0 failed)
- Manually tested in canvas with fragment blocks

## Types of changes

## Checklist:

### DIFF
--- a/blocks/shared/prose2aem.js
+++ b/blocks/shared/prose2aem.js
@@ -232,6 +232,32 @@ function parseIcons(editor) {
 
 const removeEls = (els) => els.forEach((el) => el.remove());
 
+function convertFragmentUrlsToRelative(editor) {
+  const HELIX_URL_REGEXP = /^https:\/\/.*\.(aem|hlx3?)\.(live|page)(\/|$)/;
+
+  const links = editor.querySelectorAll('a[href]');
+  links.forEach((link) => {
+    const url = link.getAttribute('href');
+
+    if (!url || !url.startsWith('https://')) {
+      return;
+    }
+
+    try {
+      const parsed = new URL(url);
+      const { host, pathname, search, hash } = parsed;
+
+      if (pathname.includes('/fragments/')
+          && host.includes('--')
+          && HELIX_URL_REGEXP.test(url)) {
+        link.setAttribute('href', `${pathname}${search}${hash}`);
+      }
+    } catch {
+      // ignore
+    }
+  });
+}
+
 /**
  * A utility to take ProseMirror formatted DOM and convert to AEM semantic markup
  * @param {HTMLElement} editor the editor dom
@@ -244,6 +270,10 @@ export default function prose2aem(editor, livePreview, isFragment = false) {
 
   editor.removeAttribute('contenteditable');
   editor.removeAttribute('translate');
+
+  if (livePreview) {
+    convertFragmentUrlsToRelative(editor);
+  }
 
   const daDiffDeletedEls = editor.querySelectorAll('da-diff-deleted');
   removeEls(daDiffDeletedEls);
--- a/blocks/shared/prose2aem.js
+++ b/blocks/shared/prose2aem.js
@@ -233,7 +233,7 @@ function parseIcons(editor) {
 const removeEls = (els) => els.forEach((el) => el.remove());
 
 function convertFragmentUrlsToRelative(editor) {
-  const EDS_FRAGMENT_URL = /^https:\/\/[^\/]*--[^\/]*\.aem\.(live|page)\/(?:.*\/)?fragments\//;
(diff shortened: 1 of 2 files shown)


blocks/shared/prose2aem.js
```diff
@@ -232,6 +232,32 @@ function parseIcons(editor) {
 
 const removeEls = (els) => els.forEach((el) => el.remove());
 
+function convertFragmentUrlsToRelative(editor) {
+  const HELIX_URL_REGEXP = /^https:\/\/.*\.(aem|hlx3?)\.(live|page)(\/|$)/;
```
gknobloch: Do we still need the hlx3?

blocks/shared/prose2aem.js
```diff
@@ -232,6 +232,32 @@ function parseIcons(editor) {
...
+
+  const links = editor.querySelectorAll('a[href]');
+  links.forEach((link) => {
+    const url = link.getAttribute('href');
+
+    if (!url || !url.startsWith('https://')) {
```
gknobloch: The RegExp is also checking for "starts with https"

Couldn't we combine the full check with that regexp? Including testing it contains "/fragments" and includes "--"

blocks/shared/prose2aem.js
kmurugulla: Removed hlx3 support, now only checking aem.live and aem.page domains

blocks/shared/prose2aem.js
kmurugulla: Consolidated into single regex that checks https, branch pattern with --, domain, and fragments path

hannessolo:
Thanks for the PR.

While it works I do have 2 concerns:
* I think this would also re-write URLs that point to a different `aem.live` url to be a local path - so main--siteA--org and main--siteB--org would both be rewritten as local paths
* It only works for the `/fragments` sub path - good in a way because it reduces the radius of what we're touching but also not good in a way because it may well be people are loading fragments from other paths...

I think the better solution would be just allowing CORS from our preview URLs to aem.live but asking all EW customers to set that up is also not great.

kmurugulla:
Thanks for the feedback. Both concerns have been addressed:

> I think this would also re-write URLs that point to a different \`aem.live\` url to be a local path - so main--siteA--org and main--siteB--org would both be rewritten as local paths

Fixed. The code now extracts org and site from \`window.location.hash\` and only converts URLs matching the exact pattern \`--${site}--${org}.aem.\`. Cross-site URLs remain absolute.

> It only works for the \`/fragments\` sub path - good in a way because it reduces the radius of what we're touching but also not good in a way because it may well be people are loading fragments from other paths...

Expanded. The code now converts any same-site URL regardless of path, not just \`/fragments/\`. This supports fragments at any level and other reusable content paths.

blocks/shared/prose2aem.js
```diff
@@ -232,6 +232,30 @@ function parseIcons(editor) {
 
 const removeEls = (els) => els.forEach((el) => el.remove());
 
+function convertFragmentUrlsToRelative(editor) {
+  const hashParts = window.location.hash.slice(2).split('/');
+  const [org, site] = hashParts;
```
hannessolo: consider using `getPathDetails` to not duplicate logic

blocks/shared/prose2aem.js
```diff
@@ -245,6 +269,10 @@ export default function prose2aem(editor, livePreview, isFragment = false) {
   editor.removeAttribute('contenteditable');
   editor.removeAttribute('translate');
 
+  if (livePreview) {
+    convertFragmentUrlsToRelative(editor);
```
hannessolo: Maybe rename to `convertLocalUrlsToRelative` ?

hannessolo (commented):
2 minor nitpicks.

I think apart from that it's good now. I was trying to think if this would rewrite something we don't want rewritten, but I couldn't think of anything.

blocks/shared/prose2aem.js
kmurugulla: Done, now using getPathDetails() to extract org and site

blocks/shared/prose2aem.js
kmurugulla: Renamed to convertLocalUrlsToRelative

hannessolo (approved):

PR #1150, opened 2026-07-20
fix(sheet): warn on save when a column has no header, show saving indicator

hannessolo:
https://sheethdr--da-live--adobe.aem.live


usman-khalid:
@hannessolo probably out of scope for this PR, but thoughts on adding a saving indicator for page/content updates as well?

hannessolo:
> @hannessolo probably out of scope for this PR, but thoughts on adding a saving indicator for page/content updates as well?

With collab it *should* be auto-saving even if you leave the page/close the collab session now. We fixed a bug on da-collab recently to flush to da-admin when the session ends. Together with https://github.com/adobe/da-live/pull/1112 it should always be saved and up to date.

kptdobe (dismissed):

sharanyavinod (approved):

PR #1154, opened 2026-07-21
fix: restrict iframe postMessage targetOrigin instead of '*'

tripodsan:
## Summary
- Three places sent `postMessage` to plugin/BYO iframes (EW canvas panel, DA Library, Prepare actions) using `targetOrigin: '*'`, which broadcasts the message payload — including an IMS access token in two cases — to whatever origin the iframe currently shows, regardless of what it was loaded with.
- Added a small shared helper, `getUrlOrigin(url, fallback = '*')`, in `blocks/shared/utils.js`, and used it in `iframe-protocol.js`, `da-library.js`, and `da-prepare.js` to derive the target origin from the iframe's own (already-resolved) `src` at the time each message is sent, falling back to `'*'` only if `src` can't be parsed into a URL.
- No behavior change for the normal case — every `postMessage` call site here fires from (or after) the iframe's `load` event, so `iframe.src` / `target.src` is always the fully-resolved absolute URL.

## Test plan
- [x] `npx eslint` on all 4 changed files — clean.
- [x] `npm test` (unit suite) run both with and without this change (via `git stash`) — identical result both times: 1673 passed, 1 pre-existing failure in `da-title.test.js` unrelated to this change, confirming no regression.
- [ ] Manually click through the 3 affected UI flows (EW canvas BYO panel, DA Library BYO plugin, `da-prepare` custom action) in a running dev instance to confirm the handshake/postMessage round-trips still work end-to-end — not done this session.

### DIFF
--- a/blocks/shared/utils.js
+++ b/blocks/shared/utils.js
@@ -264,10 +264,12 @@ export const getSheetByName = (json, name) => {
 
 export const getFirstSheet = (json) => getSheetByIndex(json, 0);
 
-export function getUrlOrigin(url, fallback = '*') {
+export function getPostMessageTargetOrigin(url, fallback = '/') {
   try {
     return new URL(url, window.location.href).origin;
-  } catch {
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(`Could not determine postMessage target origin for "${url}"`, e);
     return fallback;
   }
 }
--- a/blocks/shared/utils.js
+++ b/blocks/shared/utils.js
@@ -264,6 +264,14 @@ export const getSheetByName = (json, name) => {
 
 export const getFirstSheet = (json) => getSheetByIndex(json, 0);
 
+export function getUrlOrigin(url, fallback = '*') {
+  try {
+    return new URL(url, window.location.href).origin;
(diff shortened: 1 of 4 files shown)


kptdobe (dismissed):

blocks/shared/utils.js
```diff
@@ -264,6 +264,14 @@ export const getSheetByName = (json, name) => {
 
 export const getFirstSheet = (json) => getSheetByIndex(json, 0);
 
+export function getUrlOrigin(url, fallback = '*') {
+  try {
+    return new URL(url, window.location.href).origin;
+  } catch {
+    return fallback;
```
kptdobe, line 273: Does the fallback has to be `*` ? If url cannot be parsed, something is weird, no ? Maybe fallback should be `window.location.href` origin in this case.

tripodsan: `window.location.href` doesn't make sense.... but I agree, using `*` is probably not a good idea either.

andreituicu (dismissed):

kptdobe (approved):

lkrapf (approved):

PR #1158, opened 2026-07-21
fix(sheet): weak-compare ETags so save-after-create is not dropped

benpeter:
## Description

Sheet saves right after a file create were dropped. The drift check bailed the save before the POST.

The check compared ETags as raw strings. admin.da.live is behind Cloudflare, which gzips the GET and returns a weak ETag (`W/"hash"`), while a create returns a strong ETag (`"hash"`) for the same content. So `W/"hash"` did not equal `"hash"`, and the check reported drift that was not there. Editing an existing sheet was unaffected, since its overwrite responses are also weak.

## What changed

Compare ETags with HTTP weak comparison. A `normalizeEtag` helper strips a leading `W/`. It runs in `StaleCheck.markSynced` (the single setter of `_lastEtag`) and on the drift-check GET ETag before the comparison. Weak and strong ETags for the same content now match.

## How Has This Been Tested?

- Unit test, red then green: strong baseline `"baseline"` vs a weak drift GET `W/"baseline"` must not drift. Full unit suite: 1675 passed.
- Live check against admin.da.live: a create returns `"hash"`, the gzipped GET returns `W/"hash"`. The weak compare treats them as equal, so the save proceeds.

## Types of changes

## Checklist:

### DIFF
--- a/blocks/sheet/utils/utils.js
+++ b/blocks/sheet/utils/utils.js
@@ -4,6 +4,11 @@ import { getNx2Api } from '../../../scripts/utils.js';
 const DEBOUNCE_TIME = 1000;
 const POLL_INTERVAL = 30000;
 
+// A gzipped response through Cloudflare carries a weak etag (W/"h") while a
+// create carries a strong etag ("h") for the same content. Strip the weak
+// prefix so the two compare equal (HTTP weak comparison).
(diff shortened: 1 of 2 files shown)


benpeter:
closing -- duplicate of #1157

PR #1161, opened 2026-07-22
fix: disable save button when the user has no write permission

kptdobe:
## Summary
- The lock icon (`is-read-only` class, driven by `_readOnly` in da-title.js) already reflected read-only state correctly, but the Save action button's `disabled` binding only checked `this.disabledText` - it ignored `_readOnly` entirely.
- A user viewing a config/sheet in read-only mode could still click Save, only discovering the failure after the round-trip to the admin API.
- `renderActions()` now disables the `save` action (with a tooltip) when `_readOnly` is true. `handleAction('save')` also bails early as a guard against a stale/bypassed disabled state.

## Test plan
- [x] `npx eslint blocks/edit/da-title/da-title.js test/unit/blocks/edit/da-title/da-title.test.js`
- [x] `npx wtr "./test/unit/blocks/edit/da-title/da-title.test.js" --node-resolve` — 50 passing
- [x] Added tests: save button `disabled` reflects permissions in `renderActions`, and `handleAction('save')` is a no-op (no fetch) when read-only

### DIFF
--- a/blocks/edit/da-title/da-title.js
+++ b/blocks/edit/da-title/da-title.js
@@ -216,6 +216,9 @@ export default class DaTitle extends LitElement {
   }
 
   async handleAction(action) {
+    // Guard against a stale/bypassed disabled state (e.g. permissions changed mid-session).
+    if (action === 'save' && this._readOnly) return;
+
     this._status = null;
     this._isSending = true;
     this._actions.open = false;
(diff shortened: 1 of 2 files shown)


kptdobe:
Branch renamed to comply with this repo's CLAUDE.md branch-naming rule (max 8 lowercase alphanumeric chars) - GitHub closed this PR as a side effect. Continued as #1162.

PR #1166, opened 2026-07-22
fix(canvas): allow dropping blocks into empty outline sections

shsteimer:
## Summary
- The outline panel's `_onSectionDragOver` bailed out entirely (`if (!sec.blocks.length) return;`) when dragging a block over a section with zero blocks, making it impossible to drop a block into an empty section.
- Added `moveBlockToSection(view, fromIndex, sectionIndex)` in `blocks/canvas/editor-utils/blocks.js`, following the same delete-then-insert-with-offset-adjustment pattern as `moveBlock`, reusing `getBlockPositions`/`getSectionStartOffset`, and validating both `fromIndex` and `sectionIndex` are in range.
- Updated `_onSectionDragOver` to show a drop indicator on the "No blocks" placeholder `<li>` (tagged `data-empty-section`) when dragging a block over an empty section, and updated `_onDrop` to dispatch `moveBlockToSection` for this case (disambiguated via a new `emptySection` flag on `_dropTarget`, since a `sectionIndex`-only drop target previously always meant "reorder whole section").
- No CSS changes needed — the existing `[data-drop-position="before"/"after"]` selectors are unscoped attribute selectors, so they already apply to the empty-section placeholder `<li>`.

## Test plan
- [x] `npm run lint` passes
- [x] `npm test` passes (1689 passed, 0 failed, 4 pre-existing skipped)
- [x] Added `test/unit/blocks/canvas/editor-utils/blocks.test.js` covering `moveBlockToSection`: moving a block into an empty section, out-of-range `fromIndex`, out-of-range `sectionIndex`
- [ ] Full browser/manual verification — NOT performed. This is a collaborative, IMS-authenticated editor requiring a real org/site; no dev server or browser session was run against it in this sandbox. Please verify drag-and-drop behavior manually before merging.

### DIFF
--- a/blocks/canvas/editor-utils/blocks.js
+++ b/blocks/canvas/editor-utils/blocks.js
@@ -108,6 +108,38 @@ export function insertBlockAtSectionStart(view, dom, sectionIndex) {
   view.dispatch(view.state.tr.insert(pos, parsed).scrollIntoView());
 }
 
+function getSectionCount(view) {
+  const { doc, schema } = view.state;
+  let count = 1;
+  doc.forEach((node) => {
+    if (node.type === schema.nodes.horizontal_rule) count += 1;
+  });
+  return count;
+}
+
+export function moveBlockToSection(view, fromIndex, sectionIndex) {
+  if (!view) return;
+  const positions = getBlockPositions(view);
+  if (fromIndex < 0 || fromIndex >= positions.length) return;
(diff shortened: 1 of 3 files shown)


shsteimer:
superseded by #1167 

since I was touching a bunch of drag/drop/re-order code there, just aded this fix in

Issue #1110, opened 2026-07-15
[EW] Outline block should also show default content


hannessolo:
Outline block should show default content
Outline block should allow inserting default content

hannessolo:
Close as duplicate of #1159

PR #1167, opened 2026-07-22
feat(canvas): surface default content in outline panel

### LINKED ISSUE #1159: [ew] add default content to outline
- allow adding default content (eg new paragraph) to outline in sidepanel
- show default content in sidepanel

shsteimer:
requires https://github.com/adobe/da-nx/pull/627 to land as well for the selection/scroll into view to work

## Summary
- The outline panel now surfaces loose page content — paragraphs, headings, lists, images, code blocks, blockquotes — as expandable "Default content" groups per section, interleaved with blocks in real document order (`parseSections` in `editor-utils.js`).
- Default-content items support drag-reorder, delete, and moving between blocks and content (both directions), mirroring the existing block move/delete behavior (`editor-utils/blocks.js`).
- Selecting a block, section, content group, or content child in the outline drives the doc/WYSIWYG selection and vice versa, via a shared `editorSelectChange`/`editorProseSelectChange` pub/sub — so the doc editor, WYSIWYG view, and outline all stay in sync.
- Content-group expand/collapse state only resets on a genuine structural edit (a reparse); selecting a node only additively expands the run it belongs to — including nodes with no row of their own, like a fresh empty paragraph from pressing Enter — and never collapses other groups. Deleting a content child keeps its own run expanded too, without needing a dedicated selection workaround.

## Why
Default content (loose paragraphs/headings/etc. not wrapped in a block) previously had no representation in the outline, making it hard to see, reorder, or delete relative to blocks. This closes that gap while keeping the outline's selection and expansion behavior predictable across edits.

fix #1159

https://outldflt--da-live--adobe.aem.live/

## Test plan
- [x] `npm run lint` passes (JS + CSS)
- [x] `npm test` passes (1779 tests, 0 failures, 4 skipped)
- [x] verified by running aem locally and testing in stage sandbox site for block collection


shsteimer:
note:

I built #1166 and #1165 along with this as part of trying to make the outline panel a bit easier to use. but haven't tested the 3 together, so sort of want to merge one at time and saw there was an open issue for this, so figured I'd start with this one.

hannessolo:
Fix #1159 #1110

hannessolo:
Thanks a lot for the PR!

I'm not sure grouping adjacent default content into a single _Default Content_ group makes sense. If we want to allow users to edit the page 100% from the layout view, we'd need to allow them to also move, add and delete default content, which means we'd have to show each paragraph/image etc separately. 

Maybe we could make the current Default Content groups that this PR adds expandable to show all the actual content, so it doesn't become a giant list?

shsteimer:
I like the idea of a root section that can be expanded or collapsed. That lends itself if we eventually want to do something similar for items in collection blocks.

will make that change and let you know.

shsteimer:
@hannessolo This probably warrants additional sets of eyes and testing before merging. Here is what this looks like

collapsed by default
Screenshot 2026-07-23 at 10 08 46 AM

expanded
Screenshot 2026-07-23 at 10 09 17 AM

There are several known gaps here still:
- clicking a content item syncs the content view (`scrollIntoView`), but layout view has gaps that will require da-nx changes (currently only works for images). I plan to do this in a set of follow up PRs, but if there is a strong opinion that it needs to be in this PR, let me know and I'll try to tackle it now
- keyboard nav for outline isn't great. I handled arrow keys to expand/collapse the default content sections, but didn't attempt to fix this generally. That IMHO falls outside scope of this PR
- drag handles, move/add/delete, both for the default content sections as a whole, and for individual items. I think we should do something here, but I genuinely am not sure what the right approach to this is. My thinking is we ship without it, and see how it feels to use it/get feedback, then decide what feels like the right balance. But again, if you have thoughts on what you'd like to see as a starting point, those are welcome and happy to add them in this PR if you prefer that.

hannessolo:
Nice!

- Scroll into view is fine to leave as is for now IMO
- Opening a follow up ticket for KB navigation
- IMO we should add drag handles for the default content in the section. I want to get to the point where we can do basic editing from the layout mode, and that would be a requirement. Same for deleting, since we already have the UI for that. Adding new default content can wait though IMO

sharanyavinod:
Thank you for the contribution, this looks great! Couple minor things:
- Could we have the text left aligned, similar to how its in Files?
Screenshot 2026-07-29 at 11 36 29
Screenshot 2026-07-29 at 11 36 38

- Could we have a small part of the text instead of just the type ie. Paragraph, Heading etc? Right now its very confusing to know which is which especially since they can be moved etc.

shsteimer:
@sharanyavinod addressed both your items, agree it looks way better with those. Here is how it looks now

Screenshot 2026-07-29 at 3 58 37 PM
Screenshot 2026-07-29 at 3 59 00 PM

As I was testing this more, found myself increasingly annoyed that selection wasn't consistent and expansion was super fragile to other edits because it shifts the prose mirror index of items.

Where this stands now is that if we merge https://github.com/adobe/da-nx/pull/627 with this:
- moving the cursor in layout or content view selects the appropriate outline item (expands group if needed)
- click item in cursor scrolls item into view in both layout and content mode
- on any structural change (move, delete, add) the full outline tree is rebuilt from scratch, meaning existing expansions go back to collapsed, but tries be intelligent about keeping the active group expanded

more feedback is welcome, but I _think_ this is ready to go.

sharanyavinod (approved):

hannessolo (approved):

PR #1171, opened 2026-07-23
[ew] toolbar for blocks

hannessolo:
https://blocktb--da-live--adobe.aem.live?nx=scrollan&quick-edit=scrollan

Merge with https://github.com/adobe/da-nx/pull/663

Adds:
* Toolbar for blocks
  * Edit block in modal
  * Replace block
  * Change block variant
  * Delete block
* Merges table editing into submenu
* Adds block options / menu (parity with `/edit`)
* Adds "multi-item" blocks

https://github.com/user-attachments/assets/ced40f2f-5301-43b5-a5b7-f76c560ea9da


blocks/canvas/ew-block-toolbar/ew-block-toolbar.js
```diff
@@ -0,0 +1,257 @@
...
+    if (current === '' || (this._variantOptions || []).includes(current)) {
+      picker.value = current;
+      picker.labelOverride = '';
+    } else {
+      picker.value = '';
+      picker.labelOverride = current;
+    }
+  }
+
+  _onVariantChange(e) {
```
usman-khalid, line 134: Could we add support to apply multiple variants from the toolbar? Combining variants on a block is a common use-case e.g. `cards (bento, dark)`

usman-khalid:
The scroll when applying a variant or adding an item for a multi item block doesn't seem to be working reliably. I also don't see the applied block variants in brackets in the block selector pill.

hannessolo:
> The scroll when applying a variant or adding an item for a multi item block doesn't seem to be working reliably. I also don't see the applied block variants in brackets in the block selector pill.

I think that's my mistake, you need to set also the quick-edit param, so `?nx=scrollan&quick-edit=scrollan` - updated in the PR description now. Can you still reproduce that way?

blocks/canvas/ew-block-toolbar/ew-block-toolbar.js
hannessolo: good point. We definitely need this. Might open up a follow-up though, since we'd have to make a nice multi-picker ui.

usman-khalid:
> > The scroll when applying a variant or adding an item for a multi item block doesn't seem to be working reliably. I also don't see the applied block variants in brackets in the block selector pill.
> 
> I think that's my mistake, you need to set also the quick-edit param, so `?nx=scrollan&quick-edit=scrollan` - updated in the PR description now. Can you still reproduce that way?

Ah sorry, I didn't see the linked da-nx PR. :) All good, working as expected.

shsteimer:
one issue with the edit block modal is it prevents access to the library. so if you want to use AEM Assets selector, or Icon library, or any customer picker plugins, you can't. I think for v1, this is great, maybe that can come as a follow up as well, but I suspect people will want that

blocks/canvas/ew-block-toolbar/ew-block-toolbar.js
```diff
@@ -0,0 +1,257 @@
...
+
+  _icon(name) {
+    return html`<svg aria-hidden="true" class="icon" viewBox="0 0 20 20"><use href="/img/icons/s2-icon-${name}-20-n.svg#icon"></use></svg>`;
+  }
+
+  _onEditBlock() {
+    const pos = this.view?.state.selection.from;
+    if (pos == null) return;
+    // Open the single-block editor in a modal (see ew-editor-doc.enterBlockEdit).
+    document.querySelector('ew-editor-doc')?.enterBlockEdit(pos);
```
sharanyavinod: Could you see if we can avoid such direct querying and go instead via the canvasbus?

blocks/canvas/editor-utils/selection-toolbar.js
```diff
@@ -102,12 +136,13 @@ export function createSelectionToolbarPlugin() {
         update(view) {
           const header = document.querySelector('ew-canvas-header');
           const ev = header?.editorView;
-          if (ev !== 'content' && ev !== 'split') return;
-          if (getSelectionOriginFromIframe(view.state)) return;
-          syncToolbar(view);
+          const blockEditOpen = !!document.querySelector('ew-editor-doc')?.blockEditMode;
+          if (!blockEditOpen && !TOOLBAR_EDITOR_VIEWS.has(ev)) return;
+          syncToolbar(view, ev, blockEditOpen);
```
sharanyavinod: Same here as well via the canvasbus - editorview is already available to be usable in this manner. I think I missed switching it out, sorry about that.

sharanyavinod (commented):
Very nice! Just have a couple comments. 
Also, could we maybe simplify some of the comments in the code?

sharanyavinod (approved):

usman-khalid:
This currently allows you to author just one thing per key from the block options. Once content exists in the value cell, it doesn't pop up the menu again. Can be a follow-up IMO, but it would be nice to support authoring multiple things for a given key

Screengrab from classic DA:

PR #1174, opened 2026-07-23
feat(ew): handle read-only users correctly in EW

### LINKED ISSUE #1107: [EW] Handle read-only users in EW correctly
A few issues with read-only access in Experience Workspace:
- There's no indication in the UI that a user has read-only access. We should add a lock icon like DA has today.
- a read only user can't edit in doc-mode (correct) but in canvas, they can make phantom text updates which aren't persisted via collab to the backend (partially correct, but they should not have a false impression that they can make content edits in the canvas at all).
- undo/redo buttons should not be available to the read-only user since they can't edit content
- drag and dropping an image visually appears to work but silently fails with a 403 for read-only users

usman-khalid:
Must land with https://github.com/adobe/da-nx/pull/616

Handles read-only access in EW correctly.

- Add a lock icon (and update DA to the newer s2 lock icon for consistency) in the EW canvas header.
- Remove undo/redo buttons for read-only users
- Don't render toolbars for read-only users since they can't make any updates
- Prevent read-only users from making edits that silently fail or never persist:
  - Phantom text updates in layout mode that are never persisted
  - Remove image drag & drop for read-only users, which fails silently with a 403, giving the user a false impression they could take that action

Fixes #1107 

image

image


blocks/canvas/ew-canvas-header/ew-canvas-header.css
```diff
@@ -38,6 +38,24 @@
   overflow: hidden;
 }
 
+.lock-indicator {
```
sharanyavinod: You can use the new nx-action-btn-icon here instead - https://github.com/adobe/da-nx/blob/main/nx2/styles/buttons.css

sharanyavinod:
@usman-khalid Thank you for the contribution! Code looks good, just have a minor comment.
Also, there are a couple e2es failing related to the changes introduced, so could you take a look at those as well please?

sharanyavinod (approved):

PR #1178, opened 2026-07-25
fix: filter DA asset picker to approved DM assets

actinium15:
## Summary

Filters author-tier Dynamic Media asset browsing to Approved assets by supplying a
custom Content Advisor filter schema. Applies to both the document editor and the
Canvas/Experience Workspace asset-selector surfaces.

## What DA sends to the selector

When enabled, `renderAssetSelector` receives two additional props
(`blocks/shared/aem-assets/filter-schema.js`):

- `filterSchema` — a single `AssetStatusGroup` containing a read-only `radiogroup`
  on `property=dam:assetStatus=`, with `Approved` as its only option and default value.
- `filterSchemaSource: 'hybrid-merge-deep'`

`filterSchema` is documented in
[Content Advisor properties](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/assets/content-advisor/content-advisor-properties)
as "Model that is used to configure filter properties. This is useful when you want to
limit certain filter options in Content Advisor." Consistent with that, and with the
sample in
[Content Advisor customizations](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/assets/content-advisor/content-advisor-customization),
a supplied `filterSchema` defines the filter panel. **The filter panel is therefore
reduced to the locked Asset Status group while this is enabled.**

Selector prop construction is shared by the document editor and Canvas through
`blocks/shared/aem-assets/selector-props.js`. Both retain `upload`, `collections`,
`detail-panel`, and `advisor`; effective Dynamic Media configurations also enable
`dynamic-media` on both surfaces.

`filterSchemaSource` is not covered by the public Content Advisor properties
documentation. It is set to request that the supplied schema be reconciled with any
schema the tenant already has, rather than replacing it outright. Its effect is
tenant-dependent and is not relied on for correctness — see Safety below.

## Configuration

Opt-out key: `aem.asset.dm.approvedonly`.

- Absent or `on` → filter enabled. `off` (exact value) → disabled, selector spawns with
  no custom filter schema.
- Applies only when browsing an **author-tier** repository in an **effective Dynamic
  Media** configuration. No effect on delivery-tier browsing or Author + Publish mode.
- Site configuration takes precedence over organization configuration.
- Enabled is the default, so the key need not appear in the `admin.da.live/config`
  response.

Effective Dynamic Media (`blocks/shared/aem-assets/config.js`) is any of:
`aem.asset.dm.delivery: on`, `aem.asset.smartcrop.select: on`, an
`aem.assets.prod.origin` beginning `delivery-`, or a repository ID beginning
`delivery-`.

## Safety

On the editor surface this is a discoverability improvement layered on an existing
enforced check: `blocks/edit/da-assets/da-assets.js` already blocks insertion of assets
that are not `dam:assetStatus = approved` / `dam:activationTarget = delivery` in
Author + DM mode, and shows an error panel instead. If the filter does not take effect
on a given tenant, behavior degrades to exactly the pre-PR experience.

**Known limitation:** `blocks/canvas/ew-panel-extensions/aem-assets.js` has no
equivalent insertion-time approval check, so on Canvas the filter is not backed by a
second gate.

## Behavior change beyond the filter

Canvas `getRepositoryConfig` now shares the editor's Dynamic Media predicate, which adds
the `aem.assets.prod.origin` starting with `delivery-` condition that Canvas previously
did not evaluate. For an author-tier site configured with a delivery production origin,
Canvas now builds a DM delivery URL where it previously built an author URL — aligning
Canvas with the editor.

Canvas also includes the selector's `dynamic-media` feature when Dynamic Media is
effective, matching the document editor's existing behavior and enabling the same DM
status and rendition support on both surfaces.

## Verification

- New unit coverage: default-on behavior, explicit `off` opt-out, org/site precedence,
  author vs delivery tier, and shared selector-prop construction
  (`test/unit/blocks/shared/aem-assets/`, `test/unit/blocks/canvas/ew-panel-extensions/`,
  `test/unit/blocks/edit/da-assets/`).
- Lint and unit suites pass in CI.
- Manual smoke test on an author-tier Dynamic Media tenant confirmed:
  - the locked `Asset Status: Approved` filter is displayed;
  - the AEM search request contains `property=dam:assetStatus==approved`; and
  - the selector's active feature set includes `dynamic-media`.

### DIFF
--- a/blocks/canvas/ew-panel-extensions/aem-assets.js
+++ b/blocks/canvas/ew-panel-extensions/aem-assets.js
@@ -5,7 +5,7 @@ import {
   isDynamicMediaEnabled,
   shouldFilterApprovedAssets,
 } from '../../shared/aem-assets/config.js';
-import { getApprovedOnlyFilterProps } from '../../shared/aem-assets/filter-schema.js';
+import { buildAssetSelectorProps } from '../../shared/aem-assets/selector-props.js';
 import { getExtensionsBridge } from '../editor-utils/extensions-bridge.js';
 
 const { fetchDaConfigs, getFirstSheet } = await import(`${getNx()}/utils/daConfig.js`);
@@ -122,23 +122,6 @@ function getAssetAlt(asset) {
     || '';
 }
 
-export function buildAssetSelectorProps({
-  token,
-  repoConfig,
-  onClose,
-  handleSelection,
-}) {
-  return {
-    imsToken: token,
-    repositoryId: repoConfig.repositoryId,
-    aemTierType: repoConfig.tierType,
-    featureSet: ['upload', 'collections', 'detail-panel', 'advisor'],
-    ...getApprovedOnlyFilterProps(repoConfig.approvedOnly),
-    ...(onClose && { onClose }),
-    handleSelection,
-  };
-}
-
 // ---------------------------------------------------------------------------
 // Script loader
 // ---------------------------------------------------------------------------
@@ -165,8 +148,8 @@ export async function renderAssets({ container, org, site, onClose }) {
   const { loadIms, handleSignIn } = await import(`${getNx()}/utils/ims.js`);
   const ims = await loadIms();
   if (ims?.anonymous) handleSignIn();
-  const token = ims?.accessToken?.token;
-  if (!token) return;
+  const imsToken = ims?.accessToken?.token;
+  if (!imsToken) return;
 
   const repoConfig = await getRepositoryConfig(org, site);
   if (!repoConfig) return;
@@ -189,7 +172,7 @@ export async function renderAssets({ container, org, site, onClose }) {
   };
 
   const selectorProps = buildAssetSelectorProps({
-    token,
+    imsToken,
     repoConfig,
     onClose,
     handleSelection,
--- a/blocks/canvas/ew-panel-extensions/aem-assets.js
+++ b/blocks/canvas/ew-panel-extensions/aem-assets.js
@@ -1,6 +1,11 @@
 /* eslint-disable import/no-unresolved -- importmap */
 import { DOMParser as PMDOMParser } from 'da-y-wrapper';
 import { getNx } from '../../../scripts/utils.js';
+import {
+  isDynamicMediaEnabled,
+  shouldFilterApprovedAssets,
+} from '../../shared/aem-assets/config.js';
+import { getApprovedOnlyFilterProps } from '../../shared/aem-assets/filter-schema.js';
 import { getExtensionsBridge } from '../editor-utils/extensions-bridge.js';
 
 const { fetchDaConfigs, getFirstSheet } = await import(`${getNx()}/utils/daConfig.js`);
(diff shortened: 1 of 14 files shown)


blocks/canvas/ew-panel-extensions/aem-assets.js
```diff
@@ -100,6 +122,23 @@ function getAssetAlt(asset) {
     || '';
 }
 
+export function buildAssetSelectorProps({
```
ravuthu: Can you please move this and the one from da-assets.js to a shared folder?

blocks/canvas/ew-panel-extensions/aem-assets.js
```diff
@@ -100,6 +122,23 @@ function getAssetAlt(asset) {
...
+  token,
+  repoConfig,
+  onClose,
+  handleSelection,
+}) {
+  return {
+    imsToken: token,
+    repositoryId: repoConfig.repositoryId,
+    aemTierType: repoConfig.tierType,
+    featureSet: ['upload', 'collections', 'detail-panel', 'advisor'],
```
ravuthu: On `da-assets.js`(DA doc view), we are also pushing `dynamic-media` if DM is enabled. Not sure what the advantage of it is. Wondering if we can add the same here or remove it from `da-assets.js` as well if not required?

blocks/shared/aem-assets/config.js
```diff
@@ -0,0 +1,52 @@
...
+ * @returns {boolean}
+ */
+export function shouldFilterApprovedAssets({
+  tierType,
+  isDmEnabled,
+  configuredValue,
+}) {
+  return tierType === 'author'
+    && isDmEnabled
+    && configuredValue !== 'off';
```
ravuthu, line 51: Should we set it off by default so it won't impact existing customers? So that it's not a surprise to them when this feature is moved to Production. Like to know your thoughts.

ravuthu (changes requested):
Pretty solid overall. Please let me know your thoughts on my comments.

blocks/canvas/ew-panel-extensions/aem-assets.js
actinium15: done. moved selector prop and feature-set construction into `blocks/shared/aem-assets/selector-props.js`

blocks/canvas/ew-panel-extensions/aem-assets.js
actinium15: it enables DM status and rendition support... kept `dynamic-media`, but added it to canvas when DM is enabled for consistency.

blocks/shared/aem-assets/config.js
actinium15: while I think it should be the default, but good point about surprise-element for existing users.

switched this to opt-in right now; we can obv make it opt-out in the future, but I'm not sure if there's a way to track which setups are using what configs so that we can decide if/when's the right time to flip the default - would you know?

also, if you can orient me on where to update the source for https://docs.da.live/administrators/guides/setup-aem-assets#configuration-keys I'll PR the changes to document the config there too.

ravuthu (approved):
Looks good!

hannessolo:
Hi, why is this being merged into the branch `apprasst` - do you need it on a branch for a demo?

FYI that branch is behind `main`, and we've had changes to the asset picker on `main`. If you pull recent changes into `apprasst` it's very likely there will be a conflict.

actinium15:
@ravuthu, made the `approvedOnly` config opt-out again. Please help with the merge.

ravuthu:
> why is this being merged into the branch `apprasst` - do you need it on a branch for a demo?

So that we can validate the changes of this branch on da.live, and also wanted to see if CI checks will pass

PR #971, opened 2026-05-27
fix: handle section metadata for live preview, quick-edit and experience workspace

usman-khalid:
## Description

Section metadata is not being applied in DA's built-in live preview, Quick Edit and the Experience Workspace layout mode for sites that don't have [client-side section-metadata processing](https://github.com/adobe/aem-boilerplate/commit/26af708f813fe1aef8dbd4372ee80c8e85857022), regardless of whether server side processing via helix-html-pipeline is auto-enabled (site created after May 1st, 2026) or via feature flag `features.rendering.version: 2`. All three generate their preview HTML via `prose2aem`, so we need to handle it there.

## How Has This Been Tested?

Tested on https://sm-fix--da-live--usman-khalid.aem.page/ and directly on da.live with browser overrides for all three cases (quick-edit, EW, DA live preview)

## Screenshots (if appropriate):

Before (EW):
image

After:
image

Also removes `section-metadata` from the page outline right rail in experience workspace as it's not a real block.

image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1184, opened 2026-07-28
fix: handle legacy section metadata in live preview/quick-edit correctly

usman-khalid:
#971 added section metadata handling for DA live preview & quick-edit, but it didn't handle cases where an older site isn't enrolled into `features.rendering.v2`, so if an existing site relies on `div.section-metadata` existing for any custom metadata related stuff, this breaks.

Instead of removing the `section-metadata` div after applying section metadata, hide it instead so legacy projects can still use it and do whatever custom stuff they need to.

**TESTING**

Tested on:

- https://main--da-testing-p--doe-app-svc.aem.page/drafts/allen/section-metadata-issue where this issue was reported
- On https://main--att-business--usman-khalid.aem.page/ which has server side metadata enabled.

### DIFF
--- a/blocks/shared/prose2aem.js
+++ b/blocks/shared/prose2aem.js
@@ -219,7 +219,7 @@ function applySectionMetadata(editor) {
         section.dataset[camelKey] = value;
       }
     });
-    block.remove();
+    block.style.display = 'none';
   });
 }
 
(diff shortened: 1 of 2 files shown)


usman-khalid:
TBH, I am not sure if this is the right fix or we should instead be getting customers to enable `features.rendering.v2` in these cases and adjust their code accordingly.

Issue #1144, opened 2026-07-17
[EW] Visual diff comparison


mhaack:
Authors want to compare two versions of a page using a visual comparison. For example: create a working copy of a page and, before merging it, understand the difference visually.

Source: https://jira.corp.adobe.com/browse/SITES-47369

PR #1189, opened 2026-07-30
feat(visual diff) diff view to support LOC and MSM file changes in EW

ravuthu:
## Description

Currently, the file diff view is only supported in the DA editor. Implemented the same feature for EW as well. Layout and split views will be disabled when a file has merge conflicts to prevent a broken layout for content authors.

## Related Issue

#1144 

Screenshot 2026-07-24 at 2 18 48 PM

## Types of changes

## Checklist:


blocks/edit/prose/diff/diff-utils.js
```diff
@@ -78,23 +78,35 @@ function setDiffLabelCssVars(daEditor) {
...
 
-    const daEditor = document.querySelector('da-content')?.shadowRoot
-      ?.querySelector('da-editor');
+    if (hostEl?.shadowRoot) {
+      const sheets = [...(hostEl.shadowRoot.adoptedStyleSheets || []), locSheet];
 
-    if (daEditor?.shadowRoot) {
-      const existingSheets = daEditor.shadowRoot.adoptedStyleSheets || [];
-      daEditor.shadowRoot.adoptedStyleSheets = [...existingSheets, locSheet];
+      if (hostEl.tagName === 'EW-EDITOR-DOC') {
```
sharanyavinod: I think we should avoid this pattern. Its good to share logic thats genuinely required for shared behaviour, but anything ui specific can sit in each corresponding component itself. Loading css conditionally for different pieces would make it more complex to detangle later and genuinely unnecessary - we are creating a fresh visual experience and it can stand separate.
Let's pull EW's stylesheet loading out into ew-editor-doc.js itself, owned and scoped to that component, and leave the detection/node-view/action logic in diff-utils.js where it's genuinely shared.

blocks/canvas/ew-editor-doc/ew-editor-doc-diff.css
```diff
@@ -0,0 +1,150 @@
...
+  font-weight: 600;
+
+  &:focus-visible {
+    outline: 2px solid var(--s2-blue-800);
+    outline-offset: 2px;
+  }
+}
+
+.da-diff-btn .confirm-btn::before,
+.da-diff-btn-3part .diff-accept::before {
```
sharanyavinod: Please reuse buttons as defined in https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html wherever possible

blocks/canvas/ew-editor-doc/ew-editor-doc-diff.css
```diff
@@ -0,0 +1,150 @@
...
+  color: var(--s2-green-900);
+  border: none;
+}
+
+.loc-color-overlay.loc-regional {
+  background-color: color-mix(in srgb, var(--s2-green-200) 25%, transparent) !important;
+}
+
+.loc-color-overlay.loc-langstore {
+  background-color: color-mix(in srgb, var(--s2-red-200) 25%, transparent) !important;
```
sharanyavinod: Lets avoid important as much as possible - we should instead increase specificity if needed

blocks/canvas/ew-canvas-header/ew-canvas-header.js
```diff
@@ -84,7 +86,12 @@ class EWCanvasHeader extends LitElement {
     );
   }
 
+  setEditorView(view) {
```
sharanyavinod, line 86: Why have this just to call _setEditorView?

blocks/edit/prose/diff/diff-utils.js
ravuthu: Updated

blocks/canvas/ew-editor-doc/ew-editor-doc-diff.css
ravuthu: Updated

blocks/canvas/ew-editor-doc/ew-editor-doc-diff.css
ravuthu: Updated this, but had to override the CSS to match the visuals in the screenshot.

blocks/canvas/ew-canvas-header/ew-canvas-header.js
ravuthu: setEditorView is called in `canvas.js`, and it is also used in the same file. Thought I would create this method just to follow the convention. I kept this method instead of `_setEditorView` and used it internally.

sharanyavinod: We want to avoid other consumers setting editor view like this however - all such updates need to go through the canvasbus which already provides a way to switch the editor view.

blocks/canvas/canvas.js
```diff
@@ -251,6 +251,11 @@ export default async function decorate(block) {
   syncEditorSplitLayout({ mountRoot, view: header.editorView });
   installEditorSplitDrag(mountRoot);
 
+  canvasBus.mergeConflictsState.subscribe((detail) => {
+    header.hasUnresolvedMergeConflicts = detail?.hasMergeConflicts ?? false;
+    if (header.hasUnresolvedMergeConflicts) header.setEditorView('content');
```
sharanyavinod: Same here as well, editorview should only be updated via the canvasbus

blocks/edit/prose/diff/diff-utils.js
```diff
@@ -368,6 +380,7 @@ export function getDiffClass(elName, getSchema, dispatchTransaction, { isUpstrea
           colorOverlay.style.display = 'block';
           colorOverlay.className = 'loc-tabbed-color-overlay diff-bg-upstream';
         } else if (targetTab === 'diff') {
+          colorOverlay.style.display = 'none';
```
sharanyavinod: Would be good to do this via a class instead

blocks/edit/prose/diff/diff-utils.js
```diff
@@ -78,23 +78,31 @@ function setDiffLabelCssVars(daEditor) {
   host.style.setProperty('--diff-label-upstream', `'${labels.upstream}'`);
 }
 
+function getDiffHost(view) {
```
sharanyavinod: Think this still has the same underlying shape as earlier though. Would it be worth computing the host once when each editor initializes and passing it down explicitly instead - the same way onMergeConflictsChange already gets threaded through initProse? EW already has this/this.shadowRoot available before initProse() is called, so it wouldn't need to be derived from view at all.

blocks/canvas/ew-editor-wysiwyg/ew-editor-wysiwyg.js
```diff
@@ -62,14 +62,34 @@ export class EwEditorWysiwyg extends LitElement {
...
   }
 
   disconnectedCallback() {
+    this._unsubscribeMergeConflicts?.();
     this._unsubscribeEditorActive?.();
     this._clearQuickEditRetry();
     super.disconnectedCallback();
   }
 
+  _reloadIframe() {
```
sharanyavinod: Why do we do this explicitly here? Also this raises another question - since merge conflicts are contained to doc mode and layout mode is disabled until conflicts are resolved, why do we need explicit reload and checks in wyswyg mode? Wouldn't resolving conflicts cause an update within the doc and wouldn't this in turn by default update the wyswyg mode content?

blocks/canvas/ew-canvas-header/ew-canvas-header.js
ravuthu: Make sense! Updated

blocks/canvas/canvas.js
ravuthu: Updated

blocks/edit/prose/diff/diff-utils.js
ravuthu: Updated

blocks/edit/prose/diff/diff-utils.js
ravuthu: I agree, updated

blocks/canvas/ew-editor-wysiwyg/ew-editor-wysiwyg.js
ravuthu: It is redundant. Removed it.

PR #1203, opened 2026-08-01
fix(ew): strip library-metadata from inserted blocks, read Name field

### LINKED ISSUE #1200: [ew] Block library doesn't properly handle library metadata
- Library Metadata blocks get included in the copied content, it should be excluded
- Library Metadata name field is not displayed
  - classic DA editor seems to also ignore it, but it is documented. I think both should support/use it if present

shsteimer:
## What changed
In `transformBlock` (block-library helpers), `library-metadata` is now located and removed from a block/group *before* its insertable/previewable dom is generated, so it no longer leaks into copied content. Metadata lookup also checks the preceding sibling (not just the following one and nested descendants), and its `Name` field is read and used as the block/variant name when present.

## Why
`library-metadata` blocks were showing up in the content inserted from the block library, and the metadata's documented `Name` field was silently ignored. Fixes #1200.


mhaack:
LGTM

mhaack (approved):

Issue #1199, opened 2026-08-01
[ew] Folder and file icons look too similar


hannessolo:
the difference between a folder icon and the page icon is so small - the amount of time I click the wrong one is not funny. And yes - the nature has it that there always a page and folder with the same name next to each other - I’m quite sure we can solve this to reduce the wrong-clicks to 0

PR #1214, opened 2026-08-05
feat(canvas): file icons, copy URL, and search in file explorer

### LINKED ISSUE #1215: [ew] Enhance Files panel to make it easier for authors to navigate
- copy the url for easy placement in links
- filter to make it easier to find the thing you wnat in large or deep sites
- icons next to items for glanceability

shsteimer:
## Summary
- Adds type icons (folder/page/sheet/image/etc.) to rows in the canvas side-panel file explorer, via a new shared `blocks/shared/icons.js` (extracted from `da-list-item.js` so both browsers use one source of truth).
- Adds a hover-revealed "Copy URL" button on `.html`/`.json` rows that copies the AEM preview URL (`https://main--site--org.aem.page/...`) to the clipboard, confirmed via toast.
- Adds a filename filter to the file explorer: backed by a cached, streaming site crawl (via the same `crawl()` utility `da-search` uses, but skipping content fetches) so partial matches appear as they're found instead of blocking until the whole site is scanned. Never re-crawls once cached per site.

Fix #1215 

https://fexplore--da-live--adobe.aem.live/canvas#/aem-sandbox/block-collection/drafts/shsteimer/testing

initial
Screenshot 2026-08-05 at 9 24 29 AM

filtered + hover for copy (copy is also present in the unfiltered view)

Screenshot 2026-08-05 at 9 25 25 AM


shsteimer:
Not sure if anything like this was planned, if it is feel free to ignore. UX feedback more than welcome. But more I use canvas, more I find myself wanting something like this so I can quickly find a page to link to and copy it's url. The icons may need to be re-worked, especially given #1199 but I at least find them useful when to give me a bit of visual distinction for what is what in a large tree.

mhaack:
I think this is great. Lets quickly get some UX feedback.

sharanyavinod:
Looks very nice! Couple suggestions:
- For the search box, we use something similar for Add > Prompt in the chat panel. Might nice to use that as inspiration.
Screenshot 2026-08-20 at 19 14 57

- Also, we already have a copy capability for assistant messages in the chat. Would be nice to have a similar look and feel.
Screenshot 2026-08-20 at 19 16 48

shsteimer:
@sharanyavinod appreciate the feedback. updates made:

- Search box restyle: replaced the bordered/background search input with a borderless icon+input+clear-button layout matching da-nx's chat prompt search style.
- Category filter: added an nx-picker dropdown (All/Pages/Sheets/Fragments) that narrows both the browsed tree and search results — folders are hidden once a background crawl proves they contain no matching descendant, files are narrowed directly by extension/path convention.
- Copy button restyle: switched to the correct clipboard icon (s2-icon-paste, not s2-icon-copy), resized/recolored to match chat's copy button, and replaced the toast confirmation with the same CSS-only checkmark-swap-on-focus pattern chat uses.
- Row hover titles: copyable rows (pages/sheets) show the exact AEM URL the copy button would copy; other rows (folders, images, etc.) have no title.
- Path hint fix: search-result path hints and the crawl-based folder matching are relative to the actual tree root (org/site, or a deeper permission-scoped root), not hardcoded.
- Crawl warm-up: the background site crawl now starts on first interaction with the search bar (focusing the input or opening the category picker) instead of waiting for the first keystroke/category pick, so filtering feels instant.


https://github.com/user-attachments/assets/faea3056-2dd1-40dc-875d-4c46de1b7efb

sharanyavinod:
Thank you for the updates! Looks great. Couple minor points:
- Copy button is not reachable by keyboard
- Clearing the search field should bring back focus to the input
- Wherever possible, please reuse common styles as in https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html. For eg. for the icon buttons.

sharanyavinod (approved):

PR #1223, opened 2026-08-07
fix: filter DA asset picker to approved DM assets- #1178

ravuthu:
## Summary

Filters author-tier Dynamic Media asset browsing to Approved assets by supplying a
custom Content Advisor filter schema. Applies to both the document editor and the
Canvas/Experience Workspace asset-selector surfaces.

## What DA sends to the selector

When enabled, `renderAssetSelector` receives two additional props
(`blocks/shared/aem-assets/filter-schema.js`):

- `filterSchema` — a single `AssetStatusGroup` containing a read-only `radiogroup`
  on `property=dam:assetStatus=`, with `Approved` as its only option and default value.
- `filterSchemaSource: 'hybrid-merge-deep'`

`filterSchema` is documented in
[Content Advisor properties](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/assets/content-advisor/content-advisor-properties)
as "Model that is used to configure filter properties. This is useful when you want to
limit certain filter options in Content Advisor." Consistent with that, and with the
sample in
[Content Advisor customizations](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/assets/content-advisor/content-advisor-customization),
a supplied `filterSchema` defines the filter panel. **The filter panel is therefore
reduced to the locked Asset Status group while this is enabled.**

Selector prop construction is shared by the document editor and Canvas through
`blocks/shared/aem-assets/selector-props.js`. Both retain `upload`, `collections`,
`detail-panel`, and `advisor`; effective Dynamic Media configurations also enable
`dynamic-media` on both surfaces.

`filterSchemaSource` is not covered by the public Content Advisor properties
documentation. It is set to request that the supplied schema be reconciled with any
schema the tenant already has, rather than replacing it outright. Its effect is
tenant-dependent and is not relied on for correctness — see Safety below.

## Configuration

Opt-out key: `aem.asset.dm.approvedonly`.

- Absent or `on` → filter enabled. `off` (exact value) → disabled, selector spawns with
  no custom filter schema.
- Applies only when browsing an **author-tier** repository in an **effective Dynamic
  Media** configuration. No effect on delivery-tier browsing or Author + Publish mode.
- Site configuration takes precedence over organization configuration.
- Enabled is the default, so the key need not appear in the `admin.da.live/config`
  response.

Effective Dynamic Media (`blocks/shared/aem-assets/config.js`) is any of:
`aem.asset.dm.delivery: on`, `aem.asset.smartcrop.select: on`, an
`aem.assets.prod.origin` beginning `delivery-`, or a repository ID beginning
`delivery-`.

## Safety

On the editor surface this is a discoverability improvement layered on an existing
enforced check: `blocks/edit/da-assets/da-assets.js` already blocks insertion of assets
that are not `dam:assetStatus = approved` / `dam:activationTarget = delivery` in
Author + DM mode, and shows an error panel instead. If the filter does not take effect
on a given tenant, behavior degrades to exactly the pre-PR experience.

**Known limitation:** `blocks/canvas/ew-panel-extensions/aem-assets.js` has no
equivalent insertion-time approval check, so on Canvas the filter is not backed by a
second gate.

## Behavior change beyond the filter

Canvas `getRepositoryConfig` now shares the editor's Dynamic Media predicate, which adds
the `aem.assets.prod.origin` starting with `delivery-` condition that Canvas previously
did not evaluate. For an author-tier site configured with a delivery production origin,
Canvas now builds a DM delivery URL where it previously built an author URL — aligning
Canvas with the editor.

Canvas also includes the selector's `dynamic-media` feature when Dynamic Media is
effective, matching the document editor's existing behavior and enabling the same DM
status and rendition support on both surfaces.

## Verification

- New unit coverage: default-on behavior, explicit `off` opt-out, org/site precedence,
  author vs delivery tier, and shared selector-prop construction
  (`test/unit/blocks/shared/aem-assets/`, `test/unit/blocks/canvas/ew-panel-extensions/`,
  `test/unit/blocks/edit/da-assets/`).
- Lint and unit suites pass in CI.
- Manual smoke test on an author-tier Dynamic Media tenant confirmed:
  - the locked `Asset Status: Approved` filter is displayed;
  - the AEM search request contains `property=dam:assetStatus==approved`; and
  - the selector's active feature set includes `dynamic-media`.


ravuthu:
Reference PR - https://github.com/adobe/da-live/pull/1178

@actinium15

blocks/shared/aem-assets/config.js
```diff
@@ -0,0 +1,52 @@
...
+ * @returns {boolean}
+ */
+export function shouldFilterApprovedAssets({
+  tierType,
+  isDmEnabled,
+  configuredValue,
+}) {
+  return tierType === 'author'
+    && isDmEnabled
+    && configuredValue !== 'off';
```
hannessolo, line 51: Can we make this off by default an require the value to be `on` instead, so the behaviour doesn't change for existing customers?

hannessolo (changes requested):

blocks/shared/aem-assets/config.js
hannessolo: Discussed on slack that this makes sense

hannessolo (approved):

PR #1226, opened 2026-08-10
fix(prose): render relative media_ image src from preview origin

tripodsan:
## Summary
- Documents store image `src` as a path relative to the published page (`./media_...`), which only resolves once the page is served from its preview/live origin.
- The prose editor UI runs on a different origin, so those images 404 while editing.
- `mediaBusImage.js` now rewrites the rendered `` src to the doc's `https://main--{site}--{org}.preview.da.live/media_...` origin for display only, via a `view()`/`update()` DOM patch — the underlying document node attrs (and thus the saved source) keep the relative path unchanged.
- No changes needed to `index.js` beyond the plugin wiring that was already in place, nor to `imageFocalPoint.js` (avoided a competing `nodeViews.image` registration, which would have silently broken focal-point editing).

## Test plan
- [x] Added `test/unit/blocks/edit/prose/plugins/mediaBusImage.test.js` covering the src-rewrite helper and end-to-end rendering (asserts the rendered `` src is rewritten while the doc's `attrs.src` stays relative; non-media srcs are left untouched).
- [x] Full unit suite passes (`npm test`): 1799 passed, 0 failed, 4 skipped.
- [ ] Manual check in a browser: open a doc with a relative `./media_...` image in the prose editor and confirm it renders in both an editable and a read-only (no `write` permission) session, and that the saved document source still contains the relative path.


kptdobe (approved):
Nice!

PR #1228, opened 2026-08-10
fix: serialize Playwright CI runs against shared external test sites

tripodsan:
## Summary
- Both `playwright.yml` (Da-Admin) and `playwright-hlx.yml` (Helix) run e2e tests against fixed, shared external sites (`da-sites/da-status` and `da-testautomation/da-e2e-tests`), with no isolation between concurrent CI runs and no `concurrency:` guard in either workflow.
- Investigated a failing/"hanging" run ([run 31389050902](https://github.com/adobe/da-live/actions/runs/31389050902/job/93456129238), PR #1227): 27 tests failed, all timing out waiting for the editor (`div.ProseMirror`) to load after "Create document" — across chromium/firefox/webkit and unrelated spec files, which points to a shared-backend problem rather than isolated test flakiness. `gh run list` confirmed two other PRs' Helix Playwright jobs were running concurrently against the same shared site at the same time, and the log shows cleanup steps hitting 404s consistent with racing create/delete traffic.
- Adds a `concurrency` group (keyed on `github.workflow`, `cancel-in-progress: false`) to each workflow so runs targeting the same shared site queue instead of racing each other. Different PRs' runs of the same workflow will now wait their turn rather than hammering the shared backend simultaneously.
- Tradeoff: CI can take longer end-to-end when multiple PRs are active concurrently, since Playwright runs for the same workflow now serialize repo-wide. This is a deliberate exchange of throughput for reliability against a resource that all runs share.
- **Follow-up fix found while validating**: after the concurrency fix, a subsequent run on this PR still failed one test — `preview_publish.spec.js` "Clicking Preview opens a confirmation dialog" — with `locator.click: Test timeout of 30000ms exceeded`, blocked by `#nx-toast-host`'s "Notifications" region. Root cause: `dismissAlertBanner()` (`test/e2e/utils/utils.js`) only checked once, with a bounded 3s wait, for a dismissible alert near the start of each test. The "public sandbox" org warning banner (shown for `da-sites/da-status`, the default test org) depends on an async check that can resolve later than that window, so the one-shot check missed it, and it appeared mid-test, overlapping and blocking the later preview-button click. Fixed by arming a background watcher for the rest of the test when the banner isn't visible in the initial window, so a late appearance still gets dismissed.
- Not included (potential follow-up, not done here): giving each CI run its own namespaced test folder (e.g. embedding the run ID) for further isolation; also noticed but not fixed — a few `browser.newPage()` calls in `edit.spec.js`/`delete.spec.js`/`authenticated/collab.spec.js` are never closed, and `reporter: 'html'` has no explicit `open: 'never'` (harmless under GitHub Actions' automatic `CI=true`, but worth hardening).

## Test plan
- [x] Verified both workflow YAML files still parse correctly (`python3 -c "import yaml; yaml.safe_load(open(f))"` for both).
- [x] Watched a full run on this PR to completion post-concurrency-fix: 149/152 passed, only the one alert-banner-related failure described above (down from 27 failures pre-fix).
- [ ] After merge, confirm via `gh run list --workflow=playwright-hlx.yml` that overlapping pushes/PRs queue rather than run in parallel.
- [ ] Monitor Playwright CI failure rate over the following days to confirm it drops and no run approaches the 60-minute job timeout.


kptdobe (approved):
ah... thanks. This explains why it was so hard to reproduce!

mhaack (approved):

Issue #1134, opened 2026-07-17
[EW] Files side panel: refresh files list - automatically


mhaack:
When asking the agent to create a new page, it often doesn't show up in the listing unless the whole page is refreshed. The file list should automatically update when a new page gets created.

Source: https://jira.corp.adobe.com/browse/SITES-45629

PR #1234, opened 2026-08-11
feat(ew): add refresh and create-page actions to the files panel

### LINKED ISSUE #1133: [EW] Files side panel: refresh files list - manually
The Files listing needs a refresh icon. When asking the agent to create a new page, it often doesn't show up in the listing unless the whole page is refreshed. It would be ideal to just refresh the file listing rather than the whole page.

Source: https://jira.corp.adobe.com/browse/SITES-45627

usman-khalid:
Allows refreshing the files from the right rail, plus allows you to create a new page directly from the panel instead of having to go back to the browse view.

Fixes #1133

Also added test coverage for `ew-file-explorer`.

Attempted to also implement #1134 with page creation done by the agent, but there is some nuance. [This current code](https://github.com/adobe/da-live/blob/main/blocks/canvas/ew-file-explorer/ew-file-explorer.js#L54-L66) doesn't seem to be doing anything and I attempted a couple of fixes which didn't resolve the auto-refresh after creation by agent use-case. Will take a look at this separately.

### DIFF
--- a/blocks/canvas/ew-file-explorer/ew-file-explorer.js
+++ b/blocks/canvas/ew-file-explorer/ew-file-explorer.js
@@ -45,7 +45,9 @@ function buildTree(cache, rootFullpath) {
 }
 
 const REFRESH_ICON_SRC = '/img/icons/s2-icon-refresh-20-n.svg';
-const ADD_ICON_SRC = '/img/icons/s2-icon-addcircle-20-n.svg';
+const ADD_ICON_SRC = '/img/icons/s2-icon-fileadd-20-n.svg';
+const REFRESH_ICON_HTML = `<svg aria-hidden="true" class="icon" viewBox="0 0 20 20"><use href="${REFRESH_ICON_SRC}#icon"></use></svg>`;
+const REFRESH_SPINNER_HTML = '<span class="da-loading-spinner" aria-hidden="true"></span>';
 
 class EwFileExplorer extends LitElement {
   static properties = {
@@ -224,14 +226,20 @@ class EwFileExplorer extends LitElement {
   async _onRefreshClick() {
     if (this._refreshing) return;
     this._refreshing = true;
-    if (this._headerRefreshBtn) this._headerRefreshBtn.disabled = true;
+    if (this._headerRefreshBtn) {
+      this._headerRefreshBtn.disabled = true;
+      this._headerRefreshBtn.innerHTML = REFRESH_SPINNER_HTML;
+    }
     try {
       const expandedFullpaths = [...(this._expanded ?? [])].map((key) => `/${key}`);
       const targets = new Set([this._treeRoot, ...expandedFullpaths].filter(Boolean));
       await Promise.all([...targets].map((fp) => this._refreshPath(fp)));
     } finally {
       this._refreshing = false;
-      if (this._headerRefreshBtn) this._headerRefreshBtn.disabled = false;
+      if (this._headerRefreshBtn) {
+        this._headerRefreshBtn.disabled = false;
+        this._headerRefreshBtn.innerHTML = REFRESH_ICON_HTML;
+      }
     }
   }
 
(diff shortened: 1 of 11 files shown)


blocks/shared/da-name-dialog/da-name-dialog.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  _onClose() {
+    this._nameError = false;
+  }
+
+  render() {
+    if (!this.open) return nothing;
+    return html`
+      <nx-dialog title="${this.title ?? 'New item'}" @close=${this._onClose}>
```
sharanyavinod, line 65: You can reuse styles/classes from https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html#form where possible.

blocks/canvas/ew-file-explorer/ew-file-explorer.js
```diff
@@ -202,6 +213,44 @@ class EwFileExplorer extends LitElement {
...
+      this._refreshing = false;
+      if (this._headerRefreshBtn) this._headerRefreshBtn.disabled = false;
+    }
+  }
+
+  _getHeaderRefreshButton() {
+    if (this._headerRefreshBtn) return this._headerRefreshBtn;
+    const btn = document.createElement('button');
+    btn.type = 'button';
+    btn.className = 'header-action-btn';
```
sharanyavinod: Would be nice to reuse the buttons as well from the common css, similar to forms

sharanyavinod (commented):
Thank you for the contribution, looks great! Just have a couple minor comments. 
Also 2 questions:
- It could be slightly confusing if user clicks refresh and it feels like nothing has happened. Maybe we could have a spinner or so? 
- Maybe alongside create we could also have a create and open option, so user doesnt need two clicks to navigate, especially considering create in browse directly opens the page for edit? Wdyt?

blocks/shared/da-name-dialog/da-name-dialog.js
usman-khalid: Checked and I don't think there's anything to update here, let me know if there's anything specific.

usman-khalid:
> Thank you for the contribution, looks great! Just have a couple minor comments. Also 2 questions:
> 
> * It could be slightly confusing if user clicks refresh and it feels like nothing has happened. Maybe we could have a spinner or so?

Good idea - added.

> * Maybe alongside create we could also have a create and open option, so user doesnt need two clicks to navigate, especially considering create in browse directly opens the page for edit? Wdyt?

I wanted to add this but couldn't find the right icon, but I think it makes sense in the dialog instead anyway. Updated.

image

Create and open should be the default (like browse) so pressing enter will do that.

Also updated the add icon in the file explorer to a better one.

image

blocks/canvas/ew-file-explorer/ew-file-explorer.js
usman-khalid: done.

blocks/shared/da-name-dialog/da-name-dialog.js
sharanyavinod: The inputs,error etc ie https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html#form - ideally it should be reused from there instead of base.css, so we have one centralized set of styles being used.
Apologies for the confusion, we are still in the process of moving things, but plan is to reuse this across, including for the existing link-dialog and it would be nice to use these already for the new code.

blocks/shared/da-name-dialog/da-name-dialog.css
```diff
@@ -0,0 +1,2 @@
+/* All markup here (.da-form-field, .da-input, .da-btn-*) is styled by the shared
+   base.css already adopted above — nothing component-specific to add. */
```
sharanyavinod: I think this file can be removed

blocks/shared/da-name-dialog/da-name-dialog.js
usman-khalid: Makes sense - updated.

sharanyavinod (approved):

mhaack:
@usman-khalid please check the playwright tests. Looks like they all fail on one string check.

PR #1237, opened 2026-08-12
fix(collab): stop plugin reconfigure from wiping cursor awareness

kptdobe:
## Summary
- Fixes a real bug in `blocks/edit/prose/index.js`: `applyDelayedPlugins` reconfigures ProseMirror's plugin list once lazily-loaded chunks resolve, which gives the plugins array a new identity and makes ProseMirror destroy+recreate every plugin view — including y-prosemirror's cursor plugin, whose destroy handler unconditionally nulls the local awareness `cursor` field. A user editing before that promise resolves had their collab cursor silently wiped until their next selection change. Fixed by dispatching a no-op transaction right after `updateState` to force the recreated plugin views through an update pass.
- Fixes the flaky `Collab cursors in multiple editors` e2e test: pins `hasFocus()` on both pages so cursor broadcast isn't gated on which page/tab holds real browser focus (only one can at a time), and replaces flat timeouts with deterministic waits (plus a bounded retry) on the actual yjs awareness state.
- Fixes a second, unrelated flake in `edit.spec.js`'s "Add code mark" test (reproduces on `main` too, confirmed not caused by the above): Playwright/CDP occasionally drops a single keypress in a run of counted `ArrowLeft`/`ArrowRight` presses, leaving the cursor off by one before the code-mark backticks get typed (`code` → `ode `). Now verifies the cursor actually landed where expected via ProseMirror's own selection state, and presses the same key again to close the gap if one got dropped.

## Test plan
- [x] Ran the collab e2e test 20+ times locally against the real backend; consistently passing after the focus + reconfigure fixes (was ~30% flaky before)
- [x] Ran the full `edit.spec.js` suite 9+ times locally; consistently passing after the arrow-key drift fix (previously flaky ~1-in-3 runs)
- [x] `npx eslint` clean on all changed files

### DIFF
--- a/blocks/edit/prose/index.js
+++ b/blocks/edit/prose/index.js
@@ -509,6 +509,14 @@ function applyDelayedPlugins(pluginsPromise, schema, canWrite, basePlugins) {
     // Reconfigure the view with the full plugin list
     const newState = window.view.state.reconfigure({ plugins: pluginList });
     window.view.updateState(newState);
+    // reconfigure() gives the plugins array a new identity, so ProseMirror's plugin-view
+    // diffing (keyed on array identity, not per-plugin) destroys and recreates every plugin
+    // view — including y-prosemirror's cursor plugin. Its destroy handler unconditionally
+    // nulls the local awareness 'cursor' field, and the freshly recreated view isn't given
+    // an update pass in the same cycle, so a user editing before this promise resolves has
+    // their collab cursor silently wiped until their next selection change. Dispatching a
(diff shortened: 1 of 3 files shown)


blocks/edit/prose/index.js
mhaack, line 1: Claude recently started to add long comments like this one, do we want to keep these?

kptdobe: The pb was so complicated that I thought it might be useful. But you are right, it is way too long. Shorten all comments.

PR #1239, opened 2026-08-12
fix: stop appending .html to config paths

benpeter:
`/config#/{org}/{site}` sent `/config/{org}/{site}.html` on save. That 404s, and it made nx2 probe `/ping/{org}/{site}.html`, which fails CORS.

`getExtension` had no branch for the config view, so a config path fell through to the document default and gained a `.html`. A trailing slash hid it: `/config#/{org}/{site}/` short-circuits the extension, `/config#/{org}/{site}` does not.

Three tests asserted the broken URL. The trailing-slash tests beside them already expected no extension, so the two forms disagreed.

## Testing

1805 unit tests pass, 1804 before. Reverting the one line fails exactly those three.

Against the served module, both config forms now give `https://admin.da.live/config/{org}/{site}`. `/config#/{org}/{site}.json` keeps its extension. Edit and sheet are unchanged: `/edit#/{org}/{site}/index` still gives `index.html`, `/sheet#/{org}/{site}/data` still gives `data.json`.

Not verified: a save against a live config sheet, which needs a signed-in local session.

Supersedes #1238, whose branch name broke the e2e setup.

### DIFF
--- a/blocks/shared/pathDetails.js
+++ b/blocks/shared/pathDetails.js
@@ -91,7 +91,9 @@ function getFullDetails({ editor, pathParts, ext }) {
 function getExtension(editor, name, isFolder) {
   const nameSplit = name.split('.');
   if (nameSplit.length >= 2) return nameSplit.pop();
-  if (isFolder || editor === 'browse') return null;
+  // a config path names an org or a site, not a document, and admin.da.live answers
+  // /config/{org}/{site}.html with a 404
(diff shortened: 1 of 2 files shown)


tripodsan (approved):

mhaack (approved):

benpeter:
playwright tests failling for other reasons, merging

PR #1244, opened 2026-08-13
test: replace fixed waitForTimeout sleeps with event-driven waits

kptdobe:
## Summary
- Replace post-edit "wait to ensure it's saved" fixed sleeps with the existing `waitForSave()` helper (polls the actual `POST /source` response instead of guessing a duration)
- Drop sleeps that were already redundant with the following `expect()`'s built-in polling (e.g. a 500ms wait immediately before `expect(locator).toBeVisible()`)
- Leave Y.js-websocket-stabilization-before-typing and directory-listing eventual-consistency sleeps untouched — no DOM/network signal to poll on for those

Net: 63 `waitForTimeout` calls / 149.8s of fixed sleep -> 30 calls / 57.3s across the e2e suite.

## Test plan
- [ ] `npm run lint` clean (done locally)
- [ ] Playwright CI (playwright.yml / playwright-hlx.yml) green
- [ ] Spot-check flakiness over a few runs, especially versions/copy_rename/collab specs where several sleeps were consolidated into `waitForSave()`


kptdobe:
Local test run surfaced 10 failures — all `waitForSave()` timeouts on ProseMirror doc saves (edit/versions/copy_rename/preview_publish/formatting/acl_browse). Root cause: doc content persists to da-admin via da-collab's WebSocket/Yjs sync, not a client-side `POST /source` — so the browser-visible response `waitForSave()` polls for never arrives. Only sheet.spec.js's saves are genuine client-side POST calls (jspreadsheet), which is why that one passed.

Pushed a follow-up commit reverting the doc-save sites back to fixed sleeps; kept everything else (redundant pre-assert sleep removals, auth.setup.js, delete.spec.js modal-click fix, sheet.spec.js's pre-existing waitForSave usage).

Issue #1141, opened 2026-07-17
[EW] Prepare EW canvas for HLX 6


mhaack:

mhaack:
Related PRs

* https://github.com/adobe/da-live/pull/1070
* https://github.com/adobe/da-nx/pull/571
* https://github.com/adobe/da-nx/pull/573
* https://github.com/adobe/da-live/pull/1092
* https://github.com/adobe/da-nx/pull/594
* https://github.com/adobe/da-live/pull/1102

mhaack:
Closing as done for now.

PR #1249, opened 2026-08-15
fix(canvas): route the canvas to the site's content store

benpeter:
## Description

This pull request prepares EW Canvas for HLX6. Incl. site panels like files, outline, layout editing etc.

Test together with https://github.com/adobe/da-nx/pull/672

### before

`blocks/browse/da-list` and `blocks/edit` already routed through nx2's `isHlx6`. `blocks/canvas`
did not: it built its source url against da-admin regardless of the site's store. that produced
three outcomes on a source-bus site:

- da-admin 403, shown as `Not permitted`
- da-admin 404, passed through as loadable, so the canvas was left blank over a document that
  exists
- da-admin 200 from a pre-migration copy, which was then edited and saved in the canvas, since
  `prose.js:97` named the room `${DA_ADMIN}${pathname}`

### now

- `buildSourceUrl` is async now and routes through nx2's `isHlx6`. `initProse` names the collab
  room with that url instead of rebuilding one against da-admin
- `resolveEditorDocSession` takes the ctx, not a built url. it confirms sign-in first, so an
  anonymous visit skips the `isHlx6` probe
- the source bus is read with nx2's `daFetch`. da-admin is read with da-live's own fetch, for the
  live token read and the 401 retry
- `sessionErrorFromResponse` reports a non-401/403 status with its code and `x-error`, rather than
  folding it into `Not permitted`
- `source.uploadMedia` is used across the four upload paths, and the src is taken from the
  response, per #1219 and #1227
- `mediaBusImage` renders `./media_` from `getPreviewOrigin`, the origin the canvas fetches its
  cookie from, per #1226
- `getSourceUploadContext` reads both url shapes

## Related Issue

part of #1141. the media browser half is #1066.

## Testing

live on a source-bus site and a legacy site:

- `HEAD https://api.aem.live/{org}/sites/{site}/source/canvas-test.html` was answered with 200, and
  no `Not permitted` was shown. the collab room was named with that same url and reported
  `synced: true`
- a `./media_` src is rendered from `stage-preview.da.live`, with the relative path left in the
  node attrs. an absolute src is untouched
- a `./media_` src is written on drop, paste, toolbar and wysiwyg replace. upload
  `POST https://api.aem.live/{org}/sites/{site}/media/.canvas-test/dropped.png` 200
- saved to api.aem.live, no `content.da.live` in the document, `admin.da.live` 404 for that path
  throughout
- legacy: the collab room was named on admin.da.live, and a drop was answered with 201 at
  `POST https://admin.da.live/source/{org}/{site}/.canvas-test/legacy-drop.png`

on localhost the dropped image is not rendered on a legacy site. the read-back is answered with
401, because `content.da.live/.gimme_cookie` fails with `net::ERR_FAILED` from a localhost origin,
so the FPO placeholder is not replaced. the same happens on origin/main, served on a second port,
so this predates the branch.

## Known limitations

- [not new] a save is refused on the source bus once a second editor joins the room. da-collab comma-joins
  the connections' `Authorization` and helix parses one pair, so the write is unauthenticated
- [not new] `isHlx6` caches a legacy answer in module memory for the life of the tab (`api.js:629`). a site
  migrated mid-session is read as legacy until reload
- nx2's fetcher is used for the doc-pane uploads. it snapshots the token at page load and does not
  retry a 401. `checkDoc` uses da-live's own fetcher instead
- [not new] `canWrite` is optimistic on the source bus. `status.get` would answer it, at one request per
  open

Test:
- https://ewhlx6--da-live--adobe.aem.live/canvas?nx=hlx6cfg#/exp-workspace/frescopa6/index

### DIFF
--- a/blocks/canvas/ew-editor-doc/utils/ctx.js
+++ b/blocks/canvas/ew-editor-doc/utils/ctx.js
@@ -1,8 +1,4 @@
-import { buildSourceUrl, normalizeSourcePath } from './source.js';
-
-export function sourceUrlFromEditorCtx(ctx) {
-  return buildSourceUrl(ctx?.path);
-}
+import { normalizeSourcePath } from './source.js';
 
 export function editorCtxHasOrgRepoPath(ctx) {
   const { org, repo, path } = ctx ?? {};
--- a/blocks/canvas/ew-editor-doc/utils/ctx.js
+++ b/blocks/canvas/ew-editor-doc/utils/ctx.js
@@ -1,4 +1,4 @@
-import { buildSourceUrl } from './source.js';
+import { buildSourceUrl, normalizeSourcePath } from './source.js';
 
 export function sourceUrlFromEditorCtx(ctx) {
   return buildSourceUrl(ctx?.path);
@@ -9,8 +9,9 @@ export function editorCtxHasOrgRepoPath(ctx) {
   return Boolean(org && repo && path);
 }
 
+// Stays synchronous: the render phase needs an answer without waiting on the store lookup.
 export function editorDocCanLoad(ctx) {
-  return editorCtxHasOrgRepoPath(ctx) && Boolean(sourceUrlFromEditorCtx(ctx));
+  return editorCtxHasOrgRepoPath(ctx) && Boolean(normalizeSourcePath(ctx.path));
 }
 
 export function controllerPathnameFromEditorCtx(ctx) {
--- a/blocks/canvas/canvas.js
+++ b/blocks/canvas/canvas.js
@@ -15,7 +15,6 @@ import {
   removeSplitGutter,
 } from './ew-editor-split/ew-editor-split.js';
 import { resolveEditorDocSession } from './ew-editor-doc/utils/load-editor-doc.js';
-import { sourceUrlFromEditorCtx } from './ew-editor-doc/utils/ctx.js';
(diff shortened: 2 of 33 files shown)


blocks/canvas/ew-editor-doc/utils/ctx.js
```diff
@@ -9,8 +9,9 @@ export function editorCtxHasOrgRepoPath(ctx) {
   return Boolean(org && repo && path);
 }
 
+// Stays synchronous: the render phase needs an answer without waiting on the store lookup.
 export function editorDocCanLoad(ctx) {
-  return editorCtxHasOrgRepoPath(ctx) && Boolean(sourceUrlFromEditorCtx(ctx));
```
hannessolo, line 13: I think we can remove `sourceUrlFromEditorCtx` - it looks like both usages were replaced with `normalizeSourcePath`

hannessolo (dismissed):

mhaack:
LGTM as well, as discussed we will need to add the toast for the image upload failures. Otherwise 👍

blocks/canvas/ew-editor-doc/utils/ctx.js
benpeter: > done in 8ea1ff1a

hannessolo (approved):

mhaack (approved):

PR #1263, opened 2026-08-20
feat: support multiple items for block options via slash menu, add swatch identifier

usman-khalid:
**Must land with https://github.com/adobe/da-nx/pull/682**

- Adds support for https://github.com/adobe/da-live/pull/1171#issuecomment-5346523947
- Adds a `swatch` identifier for color options so that they can be rendered with a swatch preview in the options menu.

Test URL: https://slash--da-live--adobe.aem.live/canvas?nx=swatch#/aemsites/intuit-erp/drafts/ukhalid/account-management

DEMO:

### DIFF
--- a/blocks/canvas/ew-editor-doc/slash-menu/block-options.js
+++ b/blocks/canvas/ew-editor-doc/slash-menu/block-options.js
@@ -14,6 +14,12 @@ export function normalizeForSlashMenu(str) {
   return str?.toLowerCase().trim().replace(/\s+/g, '-');
 }
 
+function isColorCode(str) {
+  const hexColorRegex = /^#([0-9A-Fa-f]{3}|[0-9A-Fa-f]{6})$/;
+  const rgbColorRegex = /^rgba?\(\s*(\d{1,3}\s*,\s*){2}\d{1,3}(\s*,\s*(0|1|0?\.\d+))?\s*\)$/;
+  return hexColorRegex.test(str) || rgbColorRegex.test(str) || str?.includes('-gradient(');
+}
+
 function buildBlockMap(data) {
   const blockMap = new Map();
   data?.forEach((item) => {
@@ -127,7 +133,7 @@ export function blockOptionItems(state, query, blockMap = store.blockMap) {
       if (!q || v.title.toLowerCase().includes(q)) {
         const id = `blockopt-val:${i}`;
         actions.set(id, { text: v.value, moveNext: false });
-        items.push({ id, label: v.title });
+        items.push({ id, label: v.title, ...(isColorCode(v.value) ? { swatch: v.value } : {}) });
       }
     });
   }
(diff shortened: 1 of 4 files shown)


hannessolo:
I like it, but tbh I'm not sure we want a slash in the middle of a line to always open the menu (for block options it definitely makes sense)

So far it was an intentional choice to only have it allowed at the start of the line, but I can be convinced otherwise - wdyt, also asking @sharanyavinod @mhaack ?

I just don't like that this happens, and having a / in text is a legit use case...


Screenshot 2026-08-20 at 16 39 06

mhaack:
> I like it, but tbh I'm not sure we want a slash in the middle of a line to always open the menu (for block options it definitely makes sense)
> 
> So far it was an intentional choice to only have it allowed at the start of the line, but I can be convinced otherwise - wdyt, also asking @sharanyavinod @mhaack ?
> 
> I just don't like that this happens, and having a / in text is a legit use case...
> 

+1 I agree. Just typing a slash in the middle of a paragraph or sentence should not trigger the menu. I suggest to trigger the menu, while in default content when a `/` is typed:
* cursor is in empty line
* cursor is on very first position of a paragraph

mhaack:
🤔 I just checked in Obsidian client (might not be a reference, but is it actually widely used) they open the slash menu every time a user types `/`.

usman-khalid:
> I like it, but tbh I'm not sure we want a slash in the middle of a line to always open the menu (for block options it definitely makes sense)
> 
> So far it was an intentional choice to only have it allowed at the start of the line, but I can be convinced otherwise - wdyt, also asking @sharanyavinod @mhaack ?
> 
> I just don't like that this happens, and having a / in text is a legit use case...
> 
> Screenshot 2026-08-20 at 16 39 06

Fair. I do think this could be a bit disruptive. Maybe we can scope it to _just_ block option scoped tables only.

> just checked in Obsidian client (might not be a reference, but is it actually widely used) they open the slash menu very time a user types `/`.

Confluence works the same way. :)

mhaack:
If the user uses a `/` in the middle of the sentence we can maybe add some delay before we show the pop-up menu. So, at least if you’re just writing a normal piece of text, it isn’t really that much of a nuisance.

hannessolo:
If other tools do it, let's go with it until someone complains

usman-khalid:
> If the user uses a `/` in the middle of the sentence we can maybe add some delay before we show the pop-up menu. So, at least if you’re just writing a normal piece of text, it isn’t really that much of a nuisance.

@hannessolo  @mhaack - I updated this in https://github.com/adobe/da-live/pull/1263/changes/00a8a60fc64db896561fe374a0096392088395a8 to scope this only to block options, so everything else stays the same. I don't feel strongly about this either way - let me know and I can just drop the most recent commit if we're fine with the original behaviour similar to the obsidian/confluence examples.

hannessolo (approved):

mhaack (approved):

PR #1266, opened 2026-08-20
feat: improve icon library

### LINKED ISSUE #1248: [EW] Improve icon library in the right rail
A few issues with the icon library in EW:

- There's no icon preview. For large libraries, this is problematic.
- The font is not Adobe Clean, but Arial.
- Should probably have a search.

Image

usman-khalid:
Improves icon library to add:
- icon previews
- search

**Additional Updates:**
While I was in here:
- Used shared `nx-input` (for icons panel + the block library modal) where possible
- Updated icon only buttons with `nx-action-btn-icon`
- Fixed dark mode for the block library modal

Fixes #1248

### Block Library Dark Mode

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/blocks/canvas/ew-block-library-modal/ew-block-library-modal.css
+++ b/blocks/canvas/ew-block-library-modal/ew-block-library-modal.css
@@ -17,7 +17,6 @@
   grid-template-rows: auto 1fr;
   min-height: 0;
   font-family: var(--s2-font-family);
-  color: light-dark(var(--s2-gray-900), var(--s2-gray-100));
 }
 
 .modal-header {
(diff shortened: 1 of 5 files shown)


hannessolo:
Looks great. I'm a bit confused where the dark mode changes came from - AFAIK we pinned all our pages to light mode for now.
Are you implementing dark mode just so we're ready once we allow dark mode, or are you already getting dark mode activated somewhere now?

usman-khalid:
> Are you implementing dark mode just so we're ready once we allow dark mode, or are you already getting dark mode activated somewhere now?

Yes - just made some updates so that whenever we un-pin it, it works. I was doing that for the icon library stuff and the block library modal was in the same file so fixed that as well.

hannessolo (dismissed):

sharanyavinod:
@usman-khalid Looks great, thank you! Regarding search - https://github.com/adobe/da-live/pull/1214 also introduces a search in the right panel, would be nice to have the same feel here as well.

usman-khalid:
@sharanyavinod @hannessolo implemented the same search UX as #1214.

hannessolo (approved):

Issue #1129, opened 2026-07-17
[EW] Support View Port switching in layout mode


mhaack:
In layout mode, we should have the option to select a different viewport for preview, like mobile, tablet, desktop. Similar to UE. Multiple customers have asked for this already.

Source: https://jira.corp.adobe.com/browse/SITES-47746

PR #1278, opened 2026-08-26
feat(ew): add viewport switcher for block library modal and increase modal size

### LINKED ISSUE #1209: Make Block Library pop up resizable
Would it be possible to add an option to expand the block library popup? Since it has a fixed width, the blocks render at a smaller media query, so authors might end up seeing the mobile or tablet version of that block instead of the desktop version.

usman-khalid:
Fixes #1209 

Plus adds a switcher so authors can see how blocks look in each viewport while selecting one from the library.


shsteimer:
shouldn't we also support viewport switching layout view #1129? can obv be a separate/follow up PR, but giving that in library without it in layout view seems like a logical disconnect.

usman-khalid:
> shouldn't we also support viewport switching layout view #1129? can obv be a separate/follow up PR, but giving that in library without it in layout view seems like a logical disconnect.

Yup, that can be a follow-up IMO, but if we don't want to get this in without it, I'm fine to wait until we have something for #1129 or I can add it to this PR when I have some time.

mhaack:
@usman-khalid @shsteimer https://github.com/adobe/da-live/issues/1129 should be a separate PR. We are still discussing if we even should do it or not. From UE analytics we know this feature is not really used by customers.

mhaack (approved):

PR #1282, opened 2026-08-28
feat: let canvas extensions select a section

### LINKED ISSUE #1281: Canvas extensions cannot move the selection, which blocks structural authoring plugins
### Summary

A canvas panel extension can read the document and insert into it, but it has no way to **move the selection to a known place in the document**. That one gap blocks a class of "manage the structure of this page" plugins, and it is the only thing standing between the current extension API and a working authoring tool we have prototyped.

### Context

We are prototyping author tooling for section-scoped personalization on a customer site: sections carry an audience via section metadata (`Style: audience-member, audience-silver`), and each targeted section carries a generated note. Authoring this by hand is error-prone — the audience list is an unvalidated string, and a typo silently shows the section to everyone.

We built the tooling as a **BYO panel extension**, deliberately, to avoid asking for core changes. That works well for two of three jobs:

- **Read** — the handshake in `setupIframeChannel` provides `{ org, repo, ref, path }` plus the IMS token, so the plugin fetches the page source from `admin.da.live` and reports what each section targets. This is more capable than we first assumed and worth calling out in the docs.
- **Insert** — `sendHTML` reaches `insertHTML`, which does `replaceRange(from, to, …)`, so a plugin can insert correctly-formed content through the editor and stay collab-safe.
- **Locate / edit an existing section** — not possible.

### The gap

`channel.port1.onmessage` in `blocks/canvas/ew-panel-extensions/iframe-protocol.js` handles `sendText`, `sendHTML`, `setHash`, `setHref`, `closeLibrary`, `showPanel`, `setPrompt` and `getSelection`. None of these can position the selection, and the canvas hash carries no node anchor. So a plugin cannot:

1. **Jump the canvas to a section** the way `ew-page-outline` does when you click an entry. Clicking an item in a plugin's own list of sections can't scroll the editor to it.
2. **Edit an existing section.** Because `insertHTML` replaces the current selection, the edit loop already works — but only if something can put the selection on the right node first. Without that, the only route left is writing the source via `admin.da.live` underneath a live collab session, which risks clobbering or being clobbered by the y.js document. We are not doing that, so editing is simply unavailable to plugins.

### Proposed change

Expose what the outline rail already does internally, as one action:

```js
if (action === 'selectSection') {
  const ok = editorView ? selectSection(editorView, details?.index) : false;
  channel.port1.postMessage({ action: 'selectSectionResult', details: { ok } });
  return;
}
```

backed by a `selectSection(view, sectionIndex)` helper in `editor-utils/blocks.js`, which reuses the same "split top-level nodes on `horizontal_rule`" walk that `deleteSection` and `moveSection` already use, and ends in `TextSelection` + `scrollIntoView()`.

Selecting the whole section rather than just scrolling to it is deliberate: it gives the plugin the jump **and** leaves a replaceable range under `sendHTML`, so the existing insert action covers editing with no second API.

Happy to be told this belongs somewhere else, or should be shaped differently (a node/prose-index address rather than a section index would be more general, but section index is what the existing helpers already speak).

### Not asking for

Anything that lets a plugin mutate the document without going through the editor. The value here is precisely that edits stay inside ProseMirror and collab.

keepthebyte:
Closes #1281 — please read that first for the motivation. **Opening this as a draft: it is a concrete suggestion to argue about, not a merge request.** Happy for the team to take the idea and implement it their own way.

## What

Adds one action to the canvas extension channel, `selectSection`, backed by a `selectSection(view, sectionIndex)` helper in `editor-utils/blocks.js`.

## Why

A BYO panel extension can already **read** the document (the handshake hands it `{ org, repo, ref, path }` and the IMS token, so it can fetch page source from `admin.da.live`) and **insert** into it (`sendHTML` → `insertHTML`). What it cannot do is move the selection, and that blocks two things:

1. **Jumping the canvas to a section.** A plugin listing the page's sections cannot scroll the editor to one when clicked, the way `ew-page-outline` does.
2. **Editing an existing section.** `insertHTML` does `replaceRange(from, to, …)`, so the edit loop already works — it just needs something to put the selection on the right node first. Without that the only remaining route is writing source via `admin.da.live` underneath a live collab session, which we are deliberately not doing.

Selecting the whole section rather than only scrolling to it is the load-bearing choice: it delivers the jump **and** leaves a replaceable range under the existing `sendHTML`, so editing needs no second API.

## Implementation notes

- `getSectionRanges` reuses the same "split top-level nodes on `horizontal_rule`" walk that `deleteSection` and `moveSection` already use, rather than introducing a second notion of what a section is.
- Ends in `TextSelection` + `scrollIntoView()` + `view.focus()`, mirroring what the outline path does.
- Replies on the port with `{ action: 'selectSectionResult', details: { ok } }` so a plugin can feature-detect and degrade rather than fail silently — our prototype shows an explicit "this host does not support it yet" note on timeout.
- Nothing here lets a plugin mutate the document outside the editor. Edits still go through ProseMirror and collab.

## Open questions for the team

- **Address shape.** `sectionIndex` is what the existing block/section helpers already speak, but a prose-index or node address would be more general and would survive section reordering better. Happy to change.
- **Naming / placement.** `selectSection` sits alongside the other verbs in `iframe-protocol.js`; if extension capabilities are meant to be gated or declared per plugin, this should probably go through that instead.
- **Should a read-only variant exist** (`scrollToSection`) for plugins that want to navigate without disturbing the user's selection? We did not add one because the replaceable range is the point, but the two uses are separable.

## Testing

Verified against a real customer page in a local canvas: clicking a section in a plugin panel scrolls the editor to it, and a subsequent `getSelection` → edit → `sendHTML` round-trips the section's prose HTML — preserving block content and any non-audience `Style` tokens, and updating only what the plugin intended.

This is driven by an authoring plugin we built for section-scoped personalization; happy to share it if that helps evaluate the shape.

### DIFF
--- a/blocks/canvas/editor-utils/blocks.js
+++ b/blocks/canvas/editor-utils/blocks.js
@@ -1,4 +1,4 @@
-import { DOMParser as PMDOMParser, NodeSelection } from 'da-y-wrapper';
+import { DOMParser as PMDOMParser, NodeSelection, TextSelection } from 'da-y-wrapper';
 
 const NON_BLOCK_TABLE_NAMES = new Set(['metadata', 'section metadata', 'section-metadata']);
 
@@ -221,6 +221,36 @@ export function appendBlockRow(view, tablePos, rowDom) {
   const tr = state.tr.insert(tablePos + table.nodeSize - 1, rowNode);
   tr.setSelection(NodeSelection.create(tr.doc, tablePos));
   view.dispatch(tr.scrollIntoView());
+}
+
+/** Doc range of each section, split on horizontal rules (see deleteSection). */
+export function getSectionRanges(view) {
+  const { doc, schema } = view.state;
+  const ranges = [];
+  let current = null;
+  doc.forEach((node, offset) => {
+    if (node.type === schema.nodes.horizontal_rule) {
+      current = null;
+      return;
+    }
+    if (!current) {
+      current = { from: offset, to: offset };
+      ranges.push(current);
+    }
+    current.to = offset + node.nodeSize;
+  });
+  return ranges;
+}
+
+/** Select a whole section and scroll it into view. */
+export function selectSection(view, sectionIndex) {
+  if (!view) return false;
+  const range = getSectionRanges(view)[sectionIndex];
+  if (!range) return false;
+  const { tr, doc } = view.state;
+  view.dispatch(tr.setSelection(TextSelection.create(doc, range.from, range.to)).scrollIntoView());
+  view.focus();
+  return true;
 }
 
 export function deleteSection(view, sectionIndex) {
--- a/blocks/canvas/ew-panel-extensions/iframe-protocol.js
+++ b/blocks/canvas/ew-panel-extensions/iframe-protocol.js
@@ -1,4 +1,5 @@
 import { insertText, insertHTML, getEditorSelection } from './helpers.js';
+import { selectSection } from '../editor-utils/blocks.js';
 import { getNx } from '../../../scripts/utils.js';
 import { getPostMessageTargetOrigin, isValidHref } from '../../shared/utils.js';
 
@@ -59,6 +60,12 @@ export async function setupIframeChannel({ iframe, hashState, getView, onClose }
       document.dispatchEvent(
         new CustomEvent(PANEL_EVENT.OPEN, { detail: { section: 'chat', options: { text, autoSend } } }),
       );
+    }
+
+    if (action === 'selectSection') {
+      const ok = editorView ? selectSection(editorView, details?.index) : false;
+      channel.port1.postMessage({ action: 'selectSectionResult', details: { ok } });
+      return;
     }
 
     if (action === 'getSelection') {
--- a/blocks/canvas/ew-panel-extensions/iframe-protocol.js
+++ b/blocks/canvas/ew-panel-extensions/iframe-protocol.js
@@ -1,5 +1,6 @@
 import { insertText, insertHTML, getEditorSelection } from './helpers.js';
 import { selectSection } from '../editor-utils/blocks.js';
+import { canvasBus } from '../utils/canvas-bus.js';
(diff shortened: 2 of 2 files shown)


keepthebyte:
Revised after a good challenge from @keepthebyte: **the outline can't jump to a section either — it only addresses blocks and prose items.** Its section row is a drag handle with add/delete, not a click target. That's worth taking seriously, because it means the first version of this PR invented a section-level *navigation* concept the host doesn't have.

Two things fall out, and they pull in opposite directions:

**Navigation should reuse what exists.** The outline's jump is `canvasBus.editorSelectState.emit({ blockIndex, source: 'outline' })` → `ew-editor-doc._scrollDocToBlock`. Driving the view directly, as I did first, works but goes around all of it — the outline highlight stays stale and `_broadcastSelectedNode` never runs, so the wysiwyg side never learns about the selection. So navigation is now `scrollToBlock({ blockIndex })`, emitting on the same bus with the same address. A plugin listing sections just jumps to a section's first block, which is what the user wanted anyway.

**A section range is still needed, for a different reason.** `getBlockPositions` skips `NON_BLOCK_TABLE_NAMES`, so a section's **metadata table has no blockIndex and cannot be addressed at all**. That's the case our tooling actually needs — editing which audiences a section targets means reaching its section metadata. `selectSection` stays for that, and it's what leaves a replaceable range under `sendHTML`.

So: `scrollToBlock` has precedent and should be uncontroversial; `selectSection` is the genuinely new concept, and it's for reaching metadata rather than for moving the viewport. Reviewing them separately is probably right — the first may be worth taking on its own even if the second needs more thought.

One open question I didn't want to decide unilaterally: `_scrollDocToBlock` runs for any source, but `_broadcastSelectedNode(true)` only fires for `source === 'outline'`. I emit `source: 'plugin'`, so a plugin-driven jump scrolls without broadcasting. Using `'outline'` would get the broadcast but would be a lie about provenance. Whether plugin jumps should broadcast is your call.

PR #1287, opened 2026-08-31
fix(e2e): stream stale-resource listing/deletion, raise cleanup timeout

shsteimer:
Cleanup job consistently timed out because listOldTestResources listed every page before any deletes ran, all inside the default 30s test timeout. It's now an async generator that yields items as pages are fetched, so deletes overlap with listing; test timeout raised to 5min and retries scoped down to 1 for this test.

## Description

noticed this job was consistently failing and backlog was growing

## Motivation and Context

also seems like this may impact the frequent and consistent failures with PR playright tests timing out. cautiously hoping this better cleanup will help with that.

## How Has This Been Tested?

ran https://github.com/adobe/da-live/actions/runs/33434325504 which deleted many files.

## Types of changes

## Checklist:

### DIFF
--- a/test/e2e/package.json
+++ b/test/e2e/package.json
@@ -6,7 +6,7 @@
     "test:install": "playwright install",
     "test:all": "playwright test --project=chromium --project=firefox --project=webkit --grep-invert \"Delete multiple old pages\"",
     "test:nonauth": "SKIP_AUTH=true playwright test --project=chromium tests/*.spec.js --grep-invert \"Delete multiple old pages\"",
-    "test:cleanup": "playwright test --project=chromium tests/delete.spec.js --grep \"Delete multiple old pages\"",
+    "test:cleanup": "playwright test --project=chromium tests/delete.spec.js --grep \"Delete multiple old pages\" --retries=1",
     "test:ui": "playwright test --project=chromium --ui",
     "test:debug": "playwright test --project=chromium --debug",
     "test:report": "playwright show-report"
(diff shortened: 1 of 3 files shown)


test/e2e/tests/delete.spec.js
```diff
@@ -25,6 +25,7 @@ test('Delete multiple old pages', async ({ page }, workerInfo) => {
     // only execute this test on chromium
     return;
   }
+  test.setTimeout(5 * 60 * 1000);
```
shsteimer, line 28: not sure this is really needed with other change, but also not sure it hurts anything. It should give any long action a bit more breathing room to finish the cleanup, and should be safe since the gh action already has a 30 min timeout, it was just individual test timeout of 30s causing the failures.

test/e2e/package.json
```diff
@@ -6,7 +6,7 @@
     "test:install": "playwright install",
     "test:all": "playwright test --project=chromium --project=firefox --project=webkit --grep-invert \"Delete multiple old pages\"",
     "test:nonauth": "SKIP_AUTH=true playwright test --project=chromium tests/*.spec.js --grep-invert \"Delete multiple old pages\"",
-    "test:cleanup": "playwright test --project=chromium tests/delete.spec.js --grep \"Delete multiple old pages\"",
+    "test:cleanup": "playwright test --project=chromium tests/delete.spec.js --grep \"Delete multiple old pages\" --retries=1",
```
shsteimer, line 9: when they did fail, retries also failed on the pre-conditions, hence reducing to not retry since that will always fail

sharanyavinod (approved):

PR #1302, opened 2026-09-08
fix(canvas): stop hidden doc view reverting remote edits in layout mode

mhaack:
## Summary

Two collaborative-editing fixes in the Canvas editor, found while debugging "letters vanish / typing blocked" when one user edits in **layout** (quick-edit WYSIWYG) mode while another edits the same document.

**1. Hidden doc view reverting remote edits (root cause of the revert)**
`handleCursorMove` forced `view.hasFocus = () => true` on the live Yjs-bound doc-view ProseMirror even while layout view kept it hidden. That made y-prosemirror's `_isLocalCursorInView()` report `true` for the hidden view, so on every **incoming remote** edit it re-ran its relative-selection restore. The restore dispatch momentarily diverged the PM doc from the Y fragment, and y-prosemirror's `sync-plugin` `update()` hook re-syncs PM→Y unconditionally (it is *not* gated on `isChangeOrigin`) — emitting a spurious delete that reverted the remote peer's just-typed character.

Confirmed signature: user 2 sitting idle in layout mode emitted a ~10-byte delete after every ~18-byte insert received from user 1's typing. Fix: don't force-focus / move the hidden doc view's caret from the `CURSOR_MOVE` relay handler, and drop any lingering forced-focus override when the view becomes hidden. `handleSelectionChange` / `handleNodeSelect` are untouched — they never force `hasFocus`.

**2. RELOAD burst coalescing (defense-in-depth)**
The quick-edit mini-editor posts `RELOAD` (full-body resync) whenever it can't resolve its cached `cursorOffset`; a concurrent remote edit can trigger many in a row. Debounce so a burst collapses into one refresh, and clear the pending timer on port teardown. Pairs with the da-nx drift-tolerant block lookup that removes the primary source of these RELOADs.

## Test

https://collabfx--da-live--adobe.aem.live/canvas?nx=qe-index-drift#/exp-workspace/frescopa/drafts/mhaack/coffee

## Test plan

- [ ] Two users on the same doc: user 1 in doc view typing, user 2 in layout/quick-edit mode → user 1's characters persist (no flash-and-revert).
- [ ] User 2 selecting a block/image or moving the cursor in layout mode does not block or revert user 1's typing.
- [ ] Switching from layout back to doc view lands the caret correctly.
- [ ] No RELOAD storm (main-thread stall) under concurrent editing.
- [ ] Single-user editing, undo/redo, and the selection toolbar behave as before.

### DIFF
--- a/blocks/canvas/ew-editor-wysiwyg/quick-edit-controller.js
+++ b/blocks/canvas/ew-editor-wysiwyg/quick-edit-controller.js
@@ -12,6 +12,18 @@ import { MESSAGE_TYPES } from '../utils/quick-edit-messages.js';
 
 const MUTATING_MESSAGES = new Set(['node-update', 'image-replace', 'history']);
 
+// Coalesce RELOAD bursts (each rebuilds the full body) into one refresh per window;
+// a concurrent remote edit can otherwise fire many in a row and peg the main thread.
+const RELOAD_DEBOUNCE_MS = 150;
+
+function scheduleReload(ctx) {
+  if (ctx.reloadTimer) return;
+  ctx.reloadTimer = setTimeout(() => {
+    ctx.reloadTimer = null;
+    updateDocument(ctx);
+  }, RELOAD_DEBOUNCE_MS);
+}
+
 export function createControllerOnMessage(ctx) {
   return function onMessage(e) {
     const { type, payload = {} } = e.data ?? {};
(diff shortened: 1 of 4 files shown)


blocks/canvas/ew-editor-wysiwyg/quick-edit-controller.js
```diff
@@ -12,6 +12,18 @@ import { MESSAGE_TYPES } from '../utils/quick-edit-messages.js';
 
 const MUTATING_MESSAGES = new Set(['node-update', 'image-replace', 'history']);
 
+// Coalesce RELOAD bursts (each rebuilds the full body) into one refresh per window;
+// a concurrent remote edit can otherwise fire many in a row and peg the main thread.
+const RELOAD_DEBOUNCE_MS = 150;
+
+function scheduleReload(ctx) {
```
anfibiacreativa, line 19: can we get a test on scheduleReload? fake timers, fire N RELOADs inside the window, assert updateDocument runs once. it's the whole point of the change and easy to break later.

anfibiacreativa (dismissed):
Approve but please follow up on comments

blocks/canvas/ew-editor-wysiwyg/quick-edit-controller.js
mhaack: yes added some extra test case

anfibiacreativa (approved):

usman-khalid (approved):

PR #1304, opened 2026-09-08
fix: Send to Target hangs in /canvas — resolve offer id via active editor view

sagarsane:
## Summary
- "Send to Target" in the preflight menu shows up and successfully creates/updates the Target offer in both `/edit` and `/canvas`, but in `/canvas` it hangs on "Sending to Target..." and never writes the offer id back into page metadata.
- Root cause: `blocks/edit/da-prepare/actions/target/utils.js` (shared by both `/edit`'s `da-prepare.js` and `/canvas`'s `prepare-menu.js`) reads/writes the offer id by destructuring the global `window.view` ProseMirror view. That global is only ever set by `/edit`'s classic editor (`blocks/edit/prose/index.js`). `/canvas`'s document editor exposes its view instead via `getExtensionsBridge()` (`blocks/canvas/editor-utils/extensions-bridge.js`), the same accessor already used by `ew-canvas-versions.js` and `ew-page-outline.js`.
- `setOfferId()`/`removeOfferId()` had no guard, so in `/canvas` they threw a `TypeError` right after the Target API call had already succeeded — the offer was created, but writing the offer id into the page metadata table crashed silently, leaving the dialog's status text stuck on "Sending to Target..." forever.
- Fix: add a small `getActiveView()` helper that falls back from `window.view` to `getExtensionsBridge().view`, and use it in `getOfferId`, `setOfferId`, and `removeOfferId`, guarding against a missing view instead of throwing.

## Test plan
- [x] Added a failing-then-passing unit test (`test/unit/blocks/edit/da-prepare/actions/target/utils.test.js`) reproducing the crash: `removeOfferId()` with `window.view` unset and the offer id only reachable via the canvas extensions bridge.
- [x] `npm test` — full suite passes (2039 passed, 0 failed).
- [x] `npx eslint` clean on changed files.

### DIFF
--- a/blocks/edit/da-prepare/actions/target/utils.js
+++ b/blocks/edit/da-prepare/actions/target/utils.js
@@ -1,6 +1,7 @@
 import { DOMParser as ProseParser } from 'da-y-wrapper';
 import { etcFetch, getAemSiteToken, getFirstSheet } from '../../../../shared/utils.js';
 import { getNx2Api } from '../../../../../scripts/utils.js';
+import { getExtensionsBridge } from '../../../../canvas/editor-utils/extensions-bridge.js';
 import { deleteOffer, getAccessToken, getOffer, saveOffer } from './api.js';
 
 const TARGET_CONFIG_PATH = '/.da/adobe-target.json';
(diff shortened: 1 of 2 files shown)


sagarsane:
Superseded by #1305, which pushes the same fix from a branch on adobe/da-live directly (now that I have push access) instead of from my fork. Closing this one.

PR #1311, opened 2026-09-09
[EW] Show "Create document" prompt when navigating to a non-existent URL

### LINKED ISSUE #1309: [EW] Show "Create document" prompt when navigating to a non-existent URL
Current behavior in Experience Workspace is not matching with doc editor, navigating to a URL that doesn't correspond to an existing document triggers a prompt: "Create document or Cancel" dialog. This lets authors create new documents quickly by editing the URL directly — without going through the document library.

**Expected behavior**

When a user navigates to a URL that doesn't correspond to an existing document, EW should detect this and present a clear prompt:

- Create document — creates the document at the given path and opens it for editing
- Cancel — returns the user to their previous location or the document library

Screenshot from DA doc editor:

Image

bpauli:
## What

Experience Workspace (EW) now mirrors the classic doc editor: navigating to a URL whose document does not exist shows a **Create document / Cancel** prompt, instead of silently opening a blank editor and lazy-creating the document on first edit.

Closes #1309.

## Changes

- **`blocks/canvas/ew-editor-doc/utils/load-editor-doc.js`** - `resolveEditorDocSession` returns a `notFound` signal on 404 (carrying the token and sourceUrl) rather than treating a missing document as `ok`. New `createdDocSession` helper builds the post-create session with write permissions taken from the save response.
- **`blocks/canvas/ew-not-found/ew-not-found.js`** (new) - a promise-returning dialog modeled on `blocks/edit/da-not-found`, resolving to `create` / `cancel` / `hashchange`. It builds its message with DOM nodes and `textContent` so the document name (a URL path segment) cannot inject markup.
- **`blocks/canvas/canvas.js`** - prompts on `notFound`, reusing the existing `editorLoadCount` stale-load guard as the hashchange race guard (applied after both the dialog and the save awaits). Create saves `EMPTY_DOC` and mounts with the write grant from the save response, so the author is not dropped into a read-only editor. `mountCanvasEditors` is extracted so the create and normal paths share mount logic.

## Cancel behavior

Cancel returns to the **site root / document library** (`#/org/site`), not the immediate parent folder. In EW, folder paths are not valid canvas locations (clicking a folder only expands the file-explorer tree; only a document sets the hash), so navigating to a parent folder would load it as a document, 404 again, and re-open the prompt in a loop. The site root clears the canvas to the empty state with the file-explorer panel showing the folder tree, which is EW's actual document library.

## Testing

- Unit tests for the 404 -> `notFound` contract, `createdDocSession` (write perms from the save response, read-only fallback), `libraryHashFromPath`, and all four dialog resolutions.
- Full unit suite: 2050 passed, 0 failed, 4 skipped.
- ESLint: clean.

## Out of scope

- The "Open folder" affordance (v1 is Create / Cancel only).
- Back-history / true "previous location" navigation.


blocks/canvas/ew-not-found/ew-not-found.js
mhaack, line 1: We should use the NX dialog, as described here https://github.com/adobe/da-nx/blob/main/docs/dialog.md

mhaack, line 1: Additionally, maybe let's put this into the utils folder.

mhaack (changes requested):

blocks/canvas/ew-not-found/ew-not-found.js
bpauli: Done in 6cfe971. Replaced the local `da-dialog` with da-nx's `nx-dialog` (imported from `${nx}/blocks/shared/dialog/dialog.js`), following the same wrapper pattern as `da-name-dialog`: a Lit element rendering `nx-dialog` with `slot="actions"` buttons. `showEwNotFoundDialog` keeps its promise contract (`create` | `cancel` | `hashchange`).

bpauli: Done in 6cfe971. Moved the module to `blocks/canvas/utils/ew-not-found.js` (and its unit test to `test/unit/blocks/canvas/utils/`). Updated the import in `canvas.js`.

bpauli:
@mhaack Both review comments are addressed in 6cfe971:

- The not-found prompt now uses da-nx's `nx-dialog` instead of the local `da-dialog`, following the `da-name-dialog` wrapper pattern.
- The module moved to `blocks/canvas/utils/ew-not-found.js` (unit test moved alongside).

`npm run lint` and the unit tests pass locally. I did not run the full e2e suite here (it needs a live environment and IMS auth password); flagging that so it can run in CI. Ready for another look.

mhaack (approved):
LGTM but lets maybe have @hannessolo & @sharanyavinod have a quick look as well

blocks/canvas/utils/ew-not-found.js
```diff
@@ -0,0 +1,80 @@
+import { LitElement, html } from 'da-lit';
```
sharanyavinod, line 1: Why is this file inside utils?

blocks/canvas/utils/ew-not-found.js
```diff
@@ -0,0 +1,80 @@
...
+  render() {
+    // Lit escapes the interpolated name (a URL path segment), so it can't inject
+    // markup into the message.
+    return html`
+      <nx-dialog title="Document not found" @close=${this._onClose}>
+        <p>There is no document named <b><em>${this.name}</em></b> at this path.</p>
+        <p>What would you like to do?</p>
+        <button type="button" slot="actions" class="da-btn-secondary"
+          @click=${() => this._finish('cancel')}>Cancel</button>
+        <button type="button" slot="actions" class="da-btn-primary"
```
sharanyavinod, line 65: Lets use nx-form-btn-primary instead of da-btn-primary since would ideally remove all da-* styles soon https://github.com/adobe/da-nx/blob/main/nx2/styles/form.css#L5

blocks/canvas/utils/ew-not-found.js
bpauli: @sharanyavinod see feedback from @mhaack https://github.com/adobe/da-live/pull/1311#discussion_r3978717628

blocks/canvas/utils/ew-not-found.js
bpauli: Done in 6f3a9619. Switched both actions to the nx form classes: `da-btn-primary` -> `nx-form-btn-primary` and `da-btn-secondary` -> `nx-form-btn-secondary`. The component already adopts `nx2/styles/form.css`, which defines both class sets, so the buttons render the same as before. Added a unit test that asserts the action classes; `npm test` passes (2315 tests). The e2e suite needs a live env plus an auth password, so I did not run it here.

blocks/canvas/utils/ew-not-found.js
bpauli: Context on the placement: the file started at `blocks/canvas/ew-not-found/ew-not-found.js` and was moved into `blocks/canvas/utils/` earlier in this PR at @mhaack's request (https://github.com/adobe/da-live/pull/1311#discussion_r3978717628). I am leaving it in `utils/` for now so that request stays satisfied. If you both prefer a different home (for example its own `blocks/canvas/ew-not-found/` folder), align here and I will move it in a follow-up commit.

sharanyavinod: utils contains reusable helpers and methods, nothing related to rendering. So for me, this feels like the wrong place for it. @mhaack Wdyt?

mhaack: Yes this was from my previous comment. Was thinking this is/should be re-useable on browser and maybe structured content as well.

I think @sharanyavinod is right can we please move it back, for now it is not reused. We can still move it later.

blocks/canvas/ew-not-found/ew-not-found.js
```diff
@@ -0,0 +1,80 @@
+import { LitElement, html } from 'da-lit';
```
bpauli: Done in 2bcdce21. Moved it back out of `utils/`: the module is now `blocks/canvas/ew-not-found/ew-not-found.js` and its unit test is back at `test/unit/blocks/canvas/ew-not-found/ew-not-found.test.js`, with the dynamic import in `canvas.js` updated. Location move only - the nx-dialog and nx-form button changes stay as they are. `npm test` passes.

sharanyavinod (approved):

mhaack (approved):

PR #1337, opened 2026-09-14
Rebase msm onto main, and register the panel in the EW canvas

keepthebyte:
Brings the `msm` branch up to current `main` so it can be exercised again, and
registers the panel in the Experience Workspace canvas.

Follow-up to #9, where the auth question turned out to be a docs gap rather than
a worker bug.

### Why

`msm` is 1 commit ahead of `main` and **234 behind**, last touched 2026-04-08.
It also predates the EW canvas entirely — on this branch
`blocks/canvas/editor-utils/prepare-menu.js` and `blocks/canvas/ew-editor-doc/`
do not exist — so the panel is unreachable from the editor most people are now
using.

### What's here

**1. `913a4ac` rebased onto `main` — no conflicts.** That commit is 2,379
insertions and zero deletions (all new files plus six lines in
`da-prepare.js`), so 234 commits of drift produced nothing to resolve.

**2. Register MSM in the canvas prepare menu.** PR #871 only added it to the
classic `/edit` menu; the canvas keeps a separate `OOTB_ACTIONS` list, so a
`prepare` row with `Multi-site Manager` creates a menu entry whose OOTB lookup
misses, leaving `path` undefined and rendering an empty dialog. That is exactly
what we hit on production `da.live`:

```
prepare-menu.js:  ${cmp || html`<iframe src=${this._dialogItem.path} …`}
                  -> src resolves to https://da.live/canvas, empty document
```

Both menus use the same `OOTB_ACTIONS` + `render` shape and the canvas already
imports every other action from `../../edit/da-prepare/actions/`, so this is
one entry. `msm.js` reads only `org`, `site` and `path` off `details`, all of
which the canvas supplies.

### State of the tests — please read before merging

`eslint` is clean. **17 of 73 MSM unit tests fail on current main**, and the
control matters: *the same tests all pass on the unrebased `msm` branch*, so
this is drift rather than pre-existing breakage.

The cause is `fetchDaConfigs`. It used to `daFetch` `${DA_ORIGIN}/config…`
directly and now routes through the nx2 config API. The MSM tests mock
`window.fetch` by URL pattern, so the mock no longer matches and the component
renders with empty config.

That reads as a **stale test harness rather than broken runtime**, since
`helpers/config.js` calls the public `fetchDaConfigs` whose current
implementation works — but I have not proven the panel renders, and did not
want to rewrite your tests on a guess. Happy to fix the mocks in this PR if you
confirm that is the right call, or leave it to whoever owns the feature.

### Testing note that may be useful

Branch previews on `*.aem.page` resolve to **stage IMS** (nexter maps any host
containing `--` to `stage`), while `*.aem.live` hits the `.aem.live` rule first
and gets **prod**. So `msm--da-live--adobe.aem.live` is the URL to test against
production content; the `.aem.page` equivalent cannot sign in there.


keepthebyte:
Superseded by #1338.

The `msm` branch has since been brought up to date with `main` directly, so the rebase half of this PR is redundant. #1338 keeps only the part that is still needed — the single `OOTB_ACTIONS` entry registering MSM in the canvas prepare menu — against the current branch head.

Closing to keep the diff reviewable.

PR #1346, opened 2026-09-16
feat(assets): preserve last folder between asset picker openings

### LINKED ISSUE #1345: [ew] Preserve path between subsequent opening of the AEM asset picker
**customer author feedback:** every time an author selects a new asset, the picker resets to the top-level folder instead of remembering where they last navigated — forces re-browsing (or the search bar) for every asset on a page.

**Proposal:** capture the folder of the last-selected asset (`asset.path`, `dirname`) in memory for the editor session, and pass it back to the selector as its path prop next time it opens. Only do this when `repoConfig.tierType === 'author'`, delivery-tier repos have no folder structure, so path doesn't apply.

shsteimer:
## What changed
`buildAssetSelectorProps` now passes a `path` prop to the AEM asset selector, set to the folder of the last selected asset. `rememberAssetFolder` (in `blocks/shared/aem-assets/selector-props.js`) captures that folder in `buildHandleSelection`, shared by both /edit and /canvas. Only applies to author-tier repos — delivery tier has no folder structure.

## Why
Author feedback: the picker always reopened at the DAM root, forcing re-browsing (or search) for every asset inserted into a page. Remembering the last folder for the editor session avoids that.

## Reviewer notes
State is in-memory (module-level var), scoped to the page session — resets on reload, not persisted. Only captured on actual selection; browsing-then-canceling isn't tracked, since the assets-selectors.js library exposes no folder-navigation callback.

Fixes #1345

https://lastpath--da-live--adobe.aem.live/


blocks/shared/aem-assets/selector-props.js
```diff
@@ -18,6 +18,21 @@ export function buildFeatureSet(isDmEnabled) {
   return features;
 }
 
+let lastFolderPath;
```
mhaack, line 21: `lastFolderPath` is a single module-level var, not scoped per repo. The PR description says this is "scoped to the page session — resets on reload," but `edit.js:87-89,146` shows navigation is handled via `hashchange` without a full reload, so the description doesn't match actual behavior: switching orgs/sites in the same tab carries the remembered folder over into an unrelated repo's asset picker. Affects both `/edit` and `/canvas` (both route through this shared module).

Fix: key `lastFolderPath` by `repoConfig.repositoryId` instead of a flat variable.

shsteimer: good catch, updated

PR #1347, opened 2026-09-16
feat(canvas): surface content-validation results in preflight

shsteimer:
## What
Adds a "Custom" content-validation category to canvas Preflight, sourced from a project-registered validation runner in the quick-edit iframe (adobe/da-nx counterpart).

## How
`ew-editor-wysiwyg.js` opens a dedicated MessageChannel port to the quick-edit iframe and hands it to a new `createValidationRequester` (`blocks/canvas/utils/validation.js`), which owns the RUN/RESULT protocol (request correlation, a 4s timeout, sanitize+cap of accepted items). Results broadcast via a new `canvasBus.validationResultState` channel; `preflight.js` renders them grouped by check title, same as the existing References/Content/SEO categories, and hides the category entirely if no runner is registered or no result has arrived.

## Notes
- EW/canvas-only by construction — opening Preflight from classic `/edit` still emits `validationRunRequest`/subscribes to `validationResultState`, but nothing responds since `ew-editor-wysiwyg.js` never loads there, so the category just never appears. No crash, but worth a reviewer's eye.
- Depends on the https://github.com/adobe/da-nx/pull/754, which should merge before this
- will likely merge conflict with #1325, depending on merge order.

## Testing

I created a dedicated test project at https://github.com/shsteimer/validation-demo.

the checks are implemented in https://github.com/shsteimer/validation-demo/blob/main/scripts/validation-checks.js, loaded from https://github.com/shsteimer/validation-demo/blob/main/scripts/scripts.js#L211. also requires the import map at https://github.com/shsteimer/validation-demo/blob/main/head.html#L10 to ensure the quick-edit.js loaded comes from the branch.

https://valapi--da-live--adobe.aem.live/canvas?nx=valapi#/shsteimer/validation-demo/index

https://github.com/user-attachments/assets/76a2ec2f-424d-47f4-9bcf-9882c4e40683

PR #1348, opened 2026-09-16
refactor(preflight): generalize categories into a provider framework

shsteimer:
Stacked PR 1/3 — base: `main`. Followed by #1349 (custom validation provider) and #1350 (UX port). See adobe/da-live#1347 for the original combined work this was split out of (superseded by this stack).

## What
Adds a small registry (`registerPreflightProvider`/`getPreflightProviders`) so Preflight's categories come from an ordered list of providers instead of one hardcoded OOTB set. OOTB (References/Content/SEO) becomes the first provider, resolving independently of any future provider so one slow/failed provider can't block others.

## Why
Sets up the integration point for project-specific checks and extended/product-specific checks (e.g. governance) to plug into Preflight without owning the whole action — see the next PRs in this stack.

## How
The composition point (which providers are active) moves to wherever Preflight is mounted — `prepare-menu.js` (canvas) and `da-prepare.js` (classic editor) — via a shared `providers/index.js`, not `preflight.js` itself. `expandCategory` is now uniform for every category.


shsteimer:
going to close this, and all related PRs. After chatting with @mhaack and @andreituicu, we will switch this to an experience context focused pre-flight approach

PR #1351, opened 2026-09-17
fix(ew): enlarge tool-panel dropdown text for readability

### LINKED ISSUE #1196: [ew] critical UI elements are too small
The key features - select skills, open drop down with side panels are tiny font - whoever I’m presenting too can't read these - we have space - they can overlay the rest of the UI - make them bigger fonts  as they are the key UIs to use.

yevheniia0831:
Fix #1196

## What changed
- Tool-panel dropdown (`ew-tool-panel/tool-panel.js`) bumped from default `size="s"` (12px) to `size="m"` (14px) for the Editor/Library/Extensions item list, section headers, and the "Outline ▾" trigger button itself (the `nx-picker` `size` attribute controls all three together — no way to size the opened list independently without adding new CSS hooks to the shared component).

Test URLs:
- Before: https://main--da-live--adobe.aem.live/canvas?nx=fix1196#/exp-workspace/frescopaco/drafts/severinovska/bigger-text
- After: https://fix1196--da-live--adobe.aem.live/canvas?nx=fix1196#/exp-workspace/frescopaco/drafts/severinovska/bigger-text

## Follow-up (not in this PR)
The Files search bar (`ew-file-explorer.js`/`.css`) and the Prompt popover's search bar (da-nx's `prompts.js`/`.css`) are two separate, independently hand-written implementations with no shared component behind them. We manually kept their sizing in sync for this issue, but nothing enforces that going forward. Worth considering a shared search component in `nx2/blocks/shared/` at some point — deciding its API, migrating both existing call sites, and testing both still work correctly.


blocks/canvas/ew-tool-panel/tool-panel.js
```diff
@@ -248,6 +209,8 @@ class EwToolPanel extends LitElement {
           .items=${items}
           .value=${this.activeId}
           placement="below-start"
+          size="m"
+          style="--nx-picker-section-font-size: var(--s2-component-s-regular-font-size); --nx-picker-section-line-height: var(--s2-component-s-regular-line-height); --nx-picker-section-font-weight: var(--s2-body-font-weight); --nx-picker-section-color: light-dark(var(--s2-gray-600), var(--s2-gray-700)); --nx-picker-item-color: var(--s2-gray-900);"
```
mhaack: Shouldn't this be better part of the picker component?

yevheniia0831: I'm fixing it at the moment

blocks/canvas/ew-canvas-header/ew-canvas-header.css
```diff
@@ -107,7 +108,7 @@
   border-radius: 999px;
   background-color: var(--s2-blue-800, #0265dc);
   color: #fff;
-  font-size: 10px;
+  font-size: var(--s2-component-xs-regular-font-size);
```
sharanyavinod: I think this might be a good opportunity to use the nx-segmented-btn here instead and unify the styles there. Wdyt?

blocks/canvas/ew-page-outline/ew-page-outline.css
```diff
@@ -190,7 +190,7 @@
 }
 
 .content-snippet {
-  font-size: var(--s2-body-size-xs, 0.75rem);
+  font-size: var(--s2-component-xs-regular-font-size);
```
sharanyavinod: would be s here

sharanyavinod:
Screenshot 2026-09-23 at 16 07 48

The icons feel too big to me here, maybe still nice to have them 16x16, wdyt @hannessolo?

sharanyavinod:
Screenshot 2026-09-23 at 16 09 25

The add icon is still the small variant here, maybe we can switch to the bigger one now to be consistent?

blocks/canvas/ew-panel-extensions/ew-panel-library.css
```diff
@@ -202,7 +202,7 @@
   outline: none;
   background: none;
   font-family: inherit;
-  font-size: var(--s2-body-size-xs);
+  font-size: var(--s2-component-m-regular-font-size);
```
sharanyavinod: should be s here - or we make it larger already?

yevheniia0831: @sharanyavinod, this one's already the same size as File Explorer and the Prompt popover search bars. If you'd still like it smaller, I can shrink all three together to stay consistent. What do you think?

hannessolo (changes requested):
The toolbar when selecting blocks is still smaller

See here:

Image

Image

blocks/canvas/ew-canvas-header/ew-canvas-header.js
```diff
@@ -161,28 +167,14 @@ class EWCanvasHeader extends LitElement {
 
         <div class="group group-center" part="group-center">
           ${this.authorized ? html`
-          <div class="segmented" role="group" aria-label="Editor view" part="editor-view-toggle">
```
anfibiacreativa: does nx-segmented-btn already ship on the da-nx ref this points at? we're deleting the whole working segmented implementation for an import we can't see in this diff. if it's not live on the target da-nx ref yet, the editor view toggle breaks for everyone on merge. can we confirm it's deployed first?

anfibiacreativa (changes requested):

blocks/canvas/ew-canvas-header/ew-canvas-header.js
sharanyavinod: It is, and has been live for quite some time but just not integrated yet. This PR is doing exactly that, and we would adapt other usages as well iteratively. https://main--da-nx--adobe.aem.live/docs/style-guide/style-guide.html#shared-blocks

PR #1352, opened 2026-09-17
feat(ew): source image size limit from sidekick config

### LINKED ISSUE #1127: [EW] Validate image size limits client-side before upload in EW
Currently, oversized images are not validated until publish time, causing a publish error with no early warning. The image size limit should be checked client-side on drop/upload so that users receive immediate feedback instead of discovering the issue only at publish.

In scope: hardcoded limit 4.5MB applies to hlx6; 20MB is applied to sites running on previous versions of Helix

Follow-up: Helix has defined image size limits. These limits can be exposed via the Sidekick config in the Admin API, making them accessible to EW. (XWalk already has this check as well, DA does not)

Source: https://jira.corp.adobe.com/browse/SITES-46294

dicagno:
## Description

Fixes #1127

Oversized images were only rejected at publish time, surfacing as a confusing
failure with no early warning. This validates the image size client-side, up
front on the drop/paste/replace, so authors get immediate feedback.

Source the EW image upload cap from `imageSizeLimit` (in MB) on the site's
sidekick config, clamped to the 4.5 MB backend ceiling so config can only
tighten the limit, never raise it past what the upload endpoint accepts. When
the field is absent or a config fetch fails, it falls back to the ceiling, so
validation degrades safely and stays dormant (every site sees 4.5 MB) until
Helix exposes the field.

Drop the previous hlx6-only gate so the cap applies to all sites, and report
the resolved limit in the toast. Covers all three EW entry points:
drag-and-drop, paste, and quick-edit replace (each checked per image).

## Test (choose org/site/page you want to test against)

- https://imglimit--da-live--adobe.aem.live/canvas#/org/site/page

Drag/drop (or paste) or upload/replace an image over the limit into the EW editor and confirm the
"Max image size allowed is 4.5 MB" toast appears and the upload is skipped.

## Motivation and Context

Noticeable UX impact: authors previously discovered the size problem only when
publishing failed, with no indication that the image was the cause. Catching it
at drop/paste/replace time gives clear, immediate feedback.

## How Has This Been Tested?

- manually tested in Chrome (local `aem up`)
- unit-tested

## Types of changes

## Checklist:


blocks/canvas/utils/image-upload.js
```diff
@@ -19,21 +30,36 @@ export function dataUrlByteLength(dataUrl) {
...
+// Sourced from `imageSizeLimit` on the site's sidekick config so a site can
+// tighten the cap without a code change once Helix exposes the field. The read
+// goes through `getSidekickConfig`, which memoizes per site, so this is a cheap
+// lookup after the first drop and usually already warm from the editor.
+//
+async function imageLimitMB(org, site) {
+  if (!org || !site) return MAX_IMAGE_MB;
+  try {
+    const config = await getSidekickConfig({ org, site });
+    const limit = config?.imageSizeLimit;
```
mhaack: This config property does not even exist in the sidekick config? Where is this coming from?

mhaack (changes requested):

blocks/canvas/utils/image-upload.js
dicagno: Hi @mhaack thanks for your observation.
Thats a good point I actually wanted to discuss, an open point.

I know it's not optimal, but rather a proposal: this is a future proofing because from what i found out the property is not in the contract (SidekickConfig) of the Admin API, but the requirement of the issue seems to be - at least in my understanding - to have the limit configurable per-site based on the sidekick config (exposed by the same API). And a fall back is introduced (standard 4.5MB limit) in case it's not defined.

Should we open a PR to support this property in the Admin API? Otherwise, should we skip this configurability for now, or get the value from site config?

mhaack: I suggest we spilt it. For this pull request, we can take it pretty simple. We don't care about any config, and we use hard-coded limits, which is 20 MB by default as documented on aem.live/docs/limits or 4.5MB on Helix (not documented) and done.

The EDS config value, if even configured, can be read from features attribute of the side config. But this is not really documented nor can the normal logged in user read the feature attributes.

For later we can discuss on #aem-dev if a limit should be exposed via side config or sidekick config.

blocks/canvas/utils/image-upload.js
```diff
@@ -1,15 +1,21 @@
 import { getNx2, getNx2Api } from '../../../scripts/utils.js';
 
-// The api service runs on Lambda, which caps a request at 6 MiB once base64 has
-// inflated the body. Measured 2026-08-18, the AWS edge answers 413 from 4,717,360
-// bytes up, and that 413 carries no CORS header, so the browser reads it as a
-// network failure with no status. The size is checked before the request instead.
-// da-admin took a 120 MB body in the same probe, so the cap is the source bus's.
-export const MAX_IMAGE_BYTES = 4500000;
-const MAX_IMAGE_LABEL = '4.5 MB';
+// hlx6 sites upload through the source-bus, which runs on Lambda and caps a
```
mhaack: Can we make this shorter, no need for super long comments here.

blocks/canvas/utils/image-upload.js
```diff
@@ -1,15 +1,21 @@
...
+// request at 6 MiB once base64 has inflated the body. Measured 2026-08-18, the
+// AWS edge answers 413 from 4,717,360 bytes up, and that 413 carries no CORS
+// header, so the browser reads it as a network failure with no status. Those
+// sites are held to 4.5 MB. Legacy DA sites upload through da-admin, which took
+// a 120 MB body in the same probe, so they are held to the 20 MB image limit
+// documented on aem.live/limits instead. Either way the size is checked before
+// the request so authors get feedback instead of a dead-end upload.
+const MB = 1_000_000;
```
mhaack: This could be simpler with.
```
const MB = 1_000_000;
export const HLX6_MAX_IMAGE_BYTES = 4.5 * MB;
export const MAX_IMAGE_BYTES = 20 * MB;
```

blocks/canvas/utils/image-upload.js
```diff
@@ -1,15 +1,19 @@
...
-export function isImageTooLarge(bytes) {
-  return bytes > MAX_IMAGE_BYTES;
+const MB = 1_000_000;
+export const HLX6_MAX_IMAGE_MB = 4.5;
+export const MAX_IMAGE_MB = 20;
+export const HLX6_MAX_IMAGE_BYTES = HLX6_MAX_IMAGE_MB * MB;
+export const MAX_IMAGE_BYTES = MAX_IMAGE_MB * MB;
+
+export function isImageTooLarge(bytes, limitBytes) {
+  return bytes > limitBytes;
```
mhaack: This helper is one line compare and can be remoed

blocks/canvas/utils/image-upload.js
```diff
@@ -19,21 +23,32 @@ export function dataUrlByteLength(dataUrl) {
...
+  if (!isImageTooLarge(bytes, HLX6_MAX_IMAGE_BYTES)) return false;
   const [, org, site] = (parentPath ?? '').split('/');
-  if (!org || !site) return false;
-  const { isHlx6 } = await getNx2Api();
-  if (!await isHlx6(org, site)) return false;
-  await showImageTooLarge();
+  const limitMB = await imageLimitMB(org, site);
+  if (!isImageTooLarge(bytes, limitMB * MB)) return false;
+  await showImageTooLarge(limitMB);
   return true;
```
mhaack, line 47: This can be boiled down to

```
// `parentPath` is the document's folder, `/org/site/dir`.
export async function refuseOversizedImage(bytes, parentPath) {
  // Anything at or under the strictest cap is allowed on every site, so the
  // common case skips the hlx6 probe entirely.
  if (bytes <= HLX6_MAX_IMAGE_BYTES) return false;
  const [, org, site] = (parentPath ?? '').split('/');
  const limitBytes = await imageLimitBytes(org, site);
  if (bytes <= limitBytes) return false;
  await showImageTooLarge(limitBytes / MB);
  return true;
}
```

mhaack (approved):

PR #1354, opened 2026-09-18
fix(ew): unify keyboard shortcut hint styling across side panels

mhaack:
## Summary

Since this was annoying me for a while I quickly unified the hints & keyboard shortcut display and extracted the common styles.

New:
Screenshot 2026-09-18 at 16 08 14

Screenshot 2026-09-18 at 16 08 06

- Normalized the Option/Alt modifier symbol: comments panel now uses `⌥` instead of spelling out "Option", matching the versions panel.
- Versions panel's shortcut hint now renders as a grey `<kbd>` pill (matching Spectrum's convention for keyboard shortcuts and the comments panel's existing style), instead of plain inline text.
- Extracted the duplicated `.hint`/`kbd` pill styles (versions panel, comments panel's two hints) into shared `.da-hint`/`.da-kbd` classes in `blocks/shared/styles/base.css`, reused by both panels. `comments-panel.js` now also adopts `base.css`, which it wasn't loading before.

## Test plan
- [ ] Open the comments side panel in EW and verify the shortcut hint reads `⌘ + ⌥ + M` (Mac) / `Ctrl + Alt + M` (other), styled as a grey pill.
- [ ] Open the versions side panel and verify the shortcut hint now shows `⌘ + ⌥ + S` / `Ctrl + Alt + S` in the same grey pill style as the comments panel.
- [ ] Verify the compose-form "to submit" hint in the comments panel still shows its `⌘ + Enter` pill correctly sized.

# Test
- https://kbdhint--da-live--adobe.aem.live/canvas#/exp-workspace/frescopaco/drafts/mhaack/testing


usman-khalid (approved):

yevheniia0831:
A few UX notes, mostly out of scope for the CSS but worth capturing since this PR touches the exact strings/styles:

**1. Make the submit hint contextual.** The pill currently reads "⌘ + Enter to submit" regardless of context. Suggest matching the button verb: **"Or hit ⌘ + Enter to comment"** on the new-comment form, **"Or hit ⌘ + Enter to reply"** on a reply. Reads more naturally than "submit".

**2. Emphasize the prerequisite.** In `Select content and press ⌘ + ⌥ + M to add a comment.` the styled `<kbd>` pill grabs the eye, so the real first step (select text in the doc) gets missed — and the shortcut is editor-scoped, so it does nothing until the editor is focused. Bolding **Select content** would help it stand out against the pill.

**3. Add top padding to the comments hint.** It sits flush against the content above; some spacing on top would help it read as a standalone hint. (screenshots to follow)
Screenshot 2026-09-21 at 13 10 13

**4. Minor:** shared `.da-hint` has a fixed `height: 24px` — fine for the single-line versions hint, but the longer comments hint can wrap in a narrow panel and collide with the line below. `height: auto` would let it grow.
Screenshot 2026-09-21 at 13 09 52
Screenshot 2026-09-21 at 13 10 02


Separate follow-up issue (not this PR): **no edit for a posted comment** — menu only has Delete + Get link.

mhaack:
Thanks @yevheniia0831 I addressed 1-4. Editing a comment I suggest we do a extra issue. @usman-khalid is that on your list already?

usman-khalid:
> Thanks @yevheniia0831 I addressed 1-4. Editing a comment I suggest we do a extra issue. @usman-khalid is that on your list already?

Created https://github.com/adobe/da-live/issues/1365

usman-khalid (approved):

yevheniia0831 (approved):

sharanyavinod (approved):

